Issue #26, opened 2020-03-20
Having separate functions for root and final


Marius1311:
Leander told me it may be more intutitive to have one function `find_final` and another one `find_root` instead of the combined `root_final` - what do you think?

michalk8:
Agreed, let's use `root_final` internall and create 2 wrappers for fwd- and bwd- directions.

michalk8:
Implemented via #27

PR #27, opened 2020-03-20
Find root final

michalk8:
Implements #26


michalk8:
@Marius1311 do you want to complete remove root_final and replace it with `find_root` and `find_final` or just deprecate it?

PR #74, opened 2020-04-03
Change to max

Marius1311:
This changes the way the maximum value for a recurrent state in a lineage is set. Was one before, it the maximum over all transient states in that lineage now.


Marius1311:
@michalk8 , could you have a look at this please? The change is very minor, yet, tests fail.

Marius1311:
Hi Mike, this makes sense. For the recurrent states (i.e. the endpoints), I used to manually set their probability of going to themseves to 1. This makes sense intuitively but can make plotting a bit ugly because it inflates to color scale. Think of an unlikely endpoint where the calls have max a probability of 0.3 of reaching it. Now, all values are between 0 and 0.3, expect for very few that are manually set to 1. 

To circumvent it, I changed this to set the values for the endpoints differently. In the above example, I would have set this values to 0.3, i.e. to the max among all cells which are not themselves and endpoint. I see now that this can cause some trouble. Maybe it would be better to just change the color bar in plotting , rather than to change the actual values in the Lineage object?

michalk8:
I see. I think I'd be better to change it when plotting, because this brakes many things. Question is how - e.g. through percentiles?

Marius1311:
I agree, I will revert this change and just set a maximum value when plotting.

Marius1311:
@michalk8 , I changed the lineage probabilities back to what they were before. I'm only changing the color scaling now for plotting, but I leave the actual values untouched. I don't understand why the tests are still failing.

michalk8:
`plot_lin_probs` seems to be failing, specifically:
```python
cellrank/tools/_markov_chain.py:1049: in plot_lin_probs

    max_not_one = np.max(col[mask])
```
And the error is: `ValueError: zero-size array to reduction operation maximum which has no identity`.
Can it be that the array is empty? If so, why?

Marius1311:
Does this happen for roots or final cells? If it happens for roots, are you considering a single endpoint? In that case, **all** cells will have prob. 1 of going there (by design), so the above array will be empty.

Marius1311:
Your tests work very well, I'm impressed! I just updated the PR, hope that this fixes it.

PR #79, opened 2020-04-15
Including PCCA/GPCCA to identify metastable states

Marius1311:
This is still very early in development and meant to be a discussion platform.


Marius1311:
No longer relevent, see #83

PR #83, opened 2020-04-23
GPCCA Mode

Marius1311:
This is not actually meant to be merged. It's just a very hacky preliminary way to use GPCCA.


michalk8:
This is an outdated reference and everything from here should be implemented. Closing.

Marius1311:
agreed!

Issue #86, opened 2020-04-24
Saving Paga Pie


Marius1311:
When I save a paga with pie charts with a `.pdf` extension, then the legend keywords are ignored, i.e. they are just always plotted on data: 

[15.5_backwards_paga_pie.pdf](https://github.com/theislab/cellrank/files/4529336/15.5_backwards_paga_pie.pdf)

This is not the case when I save as e.g. `.png`

michalk8:
I'm not sure why is this happening, I can't reproduce it on my machine (always get the appropriate #legends, whether it's png or pdf).
There were some issues with `paga_pie` not properly displaying legends, maybe #88 fixes it.

Marius1311:
Let's talk about thing tonight, maybe we can solve it together.

michalk8:
@Marius1311 did the `scvelo` PAGA help? If not, I will take a look when we meet, since I really can't reproduce it.

Marius1311:
I haven't tried yet but will do so later!

Marius1311:
Still the same issue, saving as a pdf ignores legend position, but this is not too major at the moment, I can live with the png.

michalk8:
This also seems like an issue I cannot really fix, since it's mpl-backend dependent, over which I have no control.

Marius1311:
Okay, let's close this for now then.

PR #88, opened 2020-04-24
Fix cl fates

### LINKED ISSUE #87: Handeling both legends in PAGA pie
Getting both legends to somehow not overlap with the data is a real challenge. What's missing is the option to position the cellrank legend (`final_states`) outside of the data, like the scvelo/scanpy legend (`clusters`). 

image
scvelo legend is set to `right`, cellrank legend is `left`. The problem is, it's not really outside the plotting area. `best` does not give me anything better:
image

If at least I could turn the `clusters` legend off, and if I were able to put the cellrank legend to the place there the scvelo legend currently is, that would already help me a lot.

michalk8:
Fixes #87 and hopefully #86


michalk8:
@Marius1311 I implemented the outside legend, you can do stuff like `'top left out'` to specify it.
Should be working for both abs. prob. legend and cluster legend in `'paga_pie'`.

michalk8:
As for #86 , let me know it it still doesn't work and I wil scour the internet for more info about this.
You're testing it on Mac, right?

Marius1311:
thanks a lot Mike! I will get back to you later about this.

Marius1311:
Legend positioning works very well now, thanks.

Marius1311:
saving as pdf still displays cluster legend on data for me, on Mac. As long as I could alternatively save in png, that would be fine but there is a problem as well: saving in png ignores the passed dpi. Please try to reproduce: save the figure as png (or anything else, just not pdf), one with really high and once with really low dpi and check whether there is any difference. 

The problem for me is: there is no way to save a paga with pie charts with correct legend positioning in high resolution.

michalk8:
Could you please post here the specific command you're trying to run + which data you're using?
I will try to replicate this, but this seems more like matplotlib issue.

Marius1311:
Sure, I'm calling like this: 

````python
# set paga and cellrank parameters
paga_kwargs={'fontsize': 15,  'node_size_scale': 2, 'frameon': False, 
              'threshold': 0.1, 'max_edge_width': 4, 'edge_width_scale': 1.5, 
             'legend_loc': 'right'}
cellrank_kwargs = {'title': f'fate probabilities at E{t_point}', 'legend_kwargs': {'loc': None}}

# plot, maybe save
if save_figure: cellrank_kwargs['save'] = f'{t_point}_paga_pie.png'
cr.pl.cluster_fates(adata_t, cluster_key='clusters', mode='paga_pie', basis="umap", 
                    **cellrank_kwargs, **paga_kwargs)
````

I will upload the data to seafile and send you a link in slack.

Issue #95, opened 2020-04-30
Lineage object mixing


michalk8:

michalk8:
Implemented via #96

PR #96, opened 2020-04-30
Lineage improvements

michalk8:
Implements #95 
TODO: write some tests for this functionality.


michalk8:
Based on the tests, the issue appears to be in `louvain` package and not the changes made
here, therefore merging:
```python
AttributeError: module 'louvain._c_louvain' has no attribute '_set_rng_seed'
```

Issue #98, opened 2020-05-04
Kernel class condition number check


Marius1311:
I realised that our matrices can sometimes be ill-conditioned, see https://blogs.mathworks.com/cleve/2017/07/17/what-is-the-condition-number-of-a-matrix/ for an intro to condition numbers.

I our kernel class, I would add an option `compupte_condn` in `compute_transition_matix`, that uses `np.linalg.cond` to compute the condition number and save it somewhere in the class. In logging level `info`, I would print it out and if it's greater than 1e-15, I would issue a warning. There is currently no easy way to do this for sparse matrices, so we need to use `T.A` for this. That's why I would make it optional and set `compute_condn` to `False` by default.

michalk8:
Do you want to do this only on the Kernels or on their combinations (+, *, ...). I'd assume for the latter as well.

Marius1311:
agreed, we need it for former and latter. That paper is exactly what we need but since this is really low priority, let's do this as simple as possible and use just numpy.

michalk8:
Implemented via #99

Marius1311:
Thanks! This is really important now that I have realised that for ill conditioned matrices, we need to adjust the tolerance of `krylov-schur`!

PR #99, opened 2020-05-05
Cond num

michalk8:
Implements #98


michalk8:
Same issue as before, this will need fixing:
```python
E               AttributeError: module 'louvain' has no attribute 'set_rng_seed'
```
Apart from that, no issues with the added functionality.

PR #103, opened 2020-05-06
GPCCA and Markov chain improvements

michalk8:
Implements the GPCCA class and fixes some issues with MC.


Marius1311:
Hi Mike, I'm just going trough this and writing down things as I go along:

- passing a title to `plot_metastable_states` so far only works for `n_cells != None` and `same_plot=True`. 
  - For `c_cells=None` and `same_plot=True`, I get E1 from below
  - For `n_cells=None` and `same_plot=False`, also E1
  - For `n_cells=30` and `same_plot=False`, I get the somewhat strange behaviour of Fig 1 from below. Note that I simply passed a list of strings from 1 to 11 as title
- `plot_coarse_T` looks really good, but I wonder whether we could move the description `stationary_distribution` over to the left, basically underneath beta in Figure 2 below (with some space) 
- is the schur form (matrix R from the code) saved anywhere? I could not find it. Would be nice to have this as `self.schur_form`. Note that it's only computed for `brandts` and `scipy`, not for krylov. 
- `plot_main_states`, is the option to show the differentiation potential already implemented? I would set it to False by default, the current definition does not make too much sense I find. 
- `plot_main_states` title passing (same_plot=False) gives E1 from below
- `g_fwd.plot_main_states(same_plot=True)` does not really seem to work yet, only rest is plotted, see Fig 3 below (Note that `same_plot=False` worked fine
` g_fwd.set_main_states` and also `g_fwd.compute_main_states()` should update `adata.obs['final_states']` and also need to write to `adata.obsm['to_final_cells']` or `adata.obsm['from_root_cells']. 
- showing the `rest` does not make sense in `g_bwd.plot_main_states(n_cells=30)` when we are not looking at probabilities, i.e. `n_cells != None`. See figure 4 below. Would prefer to turn it off in that case. 
- `g_bwd.plot_main_states(n_cells=30)`, as long as `n_cells != None`, the titles should not be ['from_x', 'from_y'] or ['to_x', 'to_y'] but simply ['x', 'y']. The prefixes only make sense if we are looking at probabilities. The same thing holds for `g_fwd.plot_metastable_states` with `same_plot=False` and `n_cells != None`. So in general, we should only include prefixes when we are looking at probabilities. 
- why is `self.main_states` and `self._main_states` always `None`? Did you want to write something there? Do you think we should have another categorical type here, with the main states (basically a restriction of `self.metastable_states` with matching color array?) In case you want to do this, I already created a utility to do it, `_get_restriction_to_main(estimator: GPCCA)` in `tools._utils`




## E1
TypeError                                 Traceback (most recent call last)
<ipython-input-50-008d28393050> in <module>
----> 1 g_fwd.plot_metastable_states(legend_loc='on data', n_cells=None, same_plot=True, title = 't')

~/Projects/cellrank/cellrank/tools/estimators/_gppca.py in plot_metastable_states(self, n_cells, lineages, cluster_key, mode, time_key, show_dp, same_plot, color_map, **kwargs)
    341                 same_plot=same_plot,
    342                 color_map=color_map,
--> 343                 **kwargs,
    344             )
    345         else:

~/Projects/cellrank/cellrank/tools/estimators/_base_estimator.py in _plot_probabilities(self, attr, error_msg, lineages, cluster_key, mode, time_key, show_dp, same_plot, color_map, **kwargs)
    591                     color_gradients=A,
    592                     color_map=color_map,
--> 593                     **kwargs,
    594                 )
    595             else:

TypeError: scatter() got multiple values for keyword argument 'title'

## Fig 1
image

## Fig 2
image

## Fig 3
image

Fig 4
image

michalk8:
- 1. should be now working as expected
- 2. I think you mean right. I prefer to have my labels centered, here's how it looks:

- 3 NYI, will do. I vaguely remember from the emails from the guys from SLEPC: can't we get the form also for krylov-schur?
- 4 Default DP set to `False`
- 5 Should be working now, not sure what the issue was. I know write the lineages to `.obsm`,
    there's still an issue with the `.obs`, will get back to it later
- 6 Ok, we're not selecting the `'rest`' when `n_cells!=None`
- 7 Done
- 8 They are set after `plot_main_states(n_cells!=None, ...)`, which brings me 5) again

Currently, the `main_states` are set using `plot_main_states`, because this function has the `n_cells=30`. If user wants to plot more cells, he just needs to call this, which I think is fine.
I suggest the following in `set_main_states` (called also by `compute_main_states`), have a fixed amount (that user cannot specify, e.g. either 50 cells or based on size of the data) so that we create the categorical variables there. Otherwise, I think it's a bit confusing now.

Marius1311:
Thanks! 

Re 2: I actually meant to have it underneath 'Ductal', i.e. the x-coordinate should match the 'Ductal label' while the y-coordinate should match the stationary distribution heatmap. So basically, a y-label for the stationary distribution

Re 3: yes, we can also get it for krylov-schur, but I guess it's not yet implemented in GPCCA so we would have to do it there first. 

Will test this later!

Marius1311:
Both `compute_main_states` as well as `set_main_states` raise

```
IndexError                                Traceback (most recent call last)
<ipython-input-177-4adf3c384261> in <module>
----> 1 g_fwd.compute_main_states(method='top_n', n_main_states=3)
      2 g_fwd.plot_main_states()

~/Projects/cellrank/cellrank/tools/estimators/_gppca.py in compute_main_states(self, method, mode, alpha, min_self_prob, n_main_states)
    565             -n_main_states:
    566         ]
--> 567         self.set_main_states(names, mode)
    568 
    569     def _select_cells(

~/Projects/cellrank/cellrank/tools/estimators/_gppca.py in set_main_states(self, names, mode)
    470         self._n_cells = None  # invalidate cache
    471         self._main_states = None
--> 472         self._lin_probs = self._meta_lin_probs[names]
    473         self._dp = entropy(self._lin_probs.X.T)
    474 

~/Projects/cellrank/cellrank/tools/_lineage.py in __getitem__(self, item)
    231                 item = (item_0, item_1)
    232 
--> 233         obj = super().__getitem__(item)
    234         if shape is not None:  # keep the resulting shape
    235             obj = obj.reshape(shape)

IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

I checked the value of `names`, it's `['Alpha', 'Beta', 'Epsilon', rest]`, so I guess this happens because `rest` is not a string?

michalk8:
The Linage names should always be strings.
I had the same issue yesterday - problem was autoreload. Try restarting the notebook.

Marius1311:
Okay, that fixed it. Another problem: `g_fwd.plot_main_states(same_plot=True)` still only shows me the 'rest', nothing else:

g_fwd.plot_main_states(same_plot=True)

image

michalk8:
I'm unable to reproduce - works fine for me. Did you check the `Lineage` object?

Marius1311:
When we remove a metastable state due to a large overlap, we need to decide how to re-distribute it's probability mass. Options are

1. re-normalize (keeping proportions)
1. re-distribute evenly (every state get's the same share of probability mass)
1. re-distribute all mass to the state it overlaps with

My current favourite is 3.

Marius1311:
Re plotting, yes, the lineage probs look fine.

`g_fwd.plot_main_states()` gives
image
fine. 

`g_fwd.plot_main_states(same_plot=True)` gives
image

michalk8:
I don't understand why it plots you only the `'rest'` lineage - I've checked the code + run it myself and
it seems to be working just fine.

Seems like there are 4 possiblities for redistribution:
- even, only overlapping
- even, all
- proportional, only overlapping
- proportional, all

i like only those option with overlapping. Question: what about the stationary/initial distribution?

Marius1311:
Re stationary distribution/initial distr: should be re-computed for the reduced set of metastable states.

Marius1311:
I looked quite a bit into how to do the reduction from metastable to main states and wrote some utilities which you can find in `cellrank.tools._utils` towards the bottom. Most importantly, there is a function `reduce_lineage`, which takes a lineage object and a set of keys (corresponding to the main states) and reduces the lineage to these states while trying to keep proportions intact. I benchmarked a couple of options and arrived at the default to use mutual information between the removed states and the remaining states to do this. 

Please have a look, I would like to switch from our current implementation to using this, it would mean we don't need the 'rest' any longer. Basically, `set_main_states` and `compute_main_states` would just call `reduce_lineage` with a set of keys and a set of parameters for how to do the reduction. 

One of the utility functions (`_row_normalize`) is a duplication I believe, we can remove that.

michalk8:
@Marius1311 small comment: rather than making functions take the lineage object, you could've just defined it as a method.

michalk8:
@Marius1311 what about the overlap when assigning the cluster labels? How do you wish to handle it (the redistribution)?

Marius1311:
good question! I was thinking about this, too. Say state A overlaps with B (in terms of top cells), and we want to remove A. I would

1. test whether applying `reduce_lineage` would redistribute most of the mass from A to B (`return_weights=True`, check how much weights B gets)
1. if indeed most of the mass is redistributed from A to B, then that looks good to me and I would make this the default behaviour. 

Maybe try this ans post the weights here, than we can discuss.

Marius1311:
Just checked your updates to `set_main_states` - works!

I think I also figured out why `plot_main_states(n_cells=None)` only gives me the rest. I tried the function on the redistributed linege object and it worked, see Fig 1. Then, I tried again without redistribution and again, only the rest, see Fig. 2. Next, I tried specifying the lineages and that worked, too. So I think it must be something internal that Volker has build in in case one state is too dominant that causes only this one state to be shown. 

It's not a major problem since I know now that specifying the lineages fixes it. If we want to do it right, I guess the simplest solution would be to have a `show_rest: bool` parameter that's by default False.

michalk8:
Ok, maybe we should ask him
Question: what's the consensus on titles? I went for capitalized titles, no capitalization for labels.

michalk8:
I'd like to merge this ASAP, I'd address these questions in separate issues:
- Only 1 metastable state (stationary distribution)
- Overlap and reduction + recomputing stat/init dist
- Tests (I will fix current ones, but will begin writing ones in a separate issues)
+ Any further problems/requests that arise

Marius1311:
Hi Mike, feel free to merge this, I have no more major comments. 

A couple of small things I realised just now:

1. we write `to_final_cells_dp` to `adata.obs` - should be `to_final_states`, but I would actually opt for not writing this attribute to `adata` at all. 
1. let's replace `final_cells` by `final_states` and `root_cells` by `root_states` everywhere consistently, i.e. including things we write to `.obsm`, like `to_final_cells`. 
1. I think only calling `compute_main_states` actually writes `root/final_states` to `adata` - calling `set_main_states` does not. Is this indended? I would prefer if just calling `set_main_states` would also write to `adata`.

Marius1311:
Feel free to fix these either before or after merging!

michalk8:
Regarding the dp, we store it so that we can load it afterwards. I'd keep it there.

PR #109, opened 2020-05-11
Move to scvelo.pl.paga, fix basis

### LINKED ISSUE #25: PAGA with Pie Charts
Leander and Volker both had similar bugs where their PAGA with pie charts looks something like this: 
image

Any idea what may be happening here?

michalk8:
Fixes #25


Marius1311:
great!

PR #127, opened 2020-05-14
Add an option to not use negative cosines

Marius1311:
Still missing:
- removing the negative cosines can make some cells absorbing. Explicitly make them absorbing in the transition matrix and warn the user about this
- make sure all rows still sum to one. raise if that's not the case.


Marius1311:
There was a problem in the transition matrix function - using a connectivity kernel and sigma_corr of 30 seems to be too high for small matrices. This tells me that sigma corr should have something to do with the median correlation in my graph.

Marius1311:
# Summary on the parameters

## `use_negative_cosines`
This removed negative cosines correlations in the velocity graph. As a result, the MC becomes reducible, if few neighbors are considered only. This can be an advantage, see Fig 1 and 2 below. In that case, the recurrent classes correspond well to the end-points, and absorption probs towards them make sense. Dealing with the reducible MC is not a problem for CFLARE, but it is for GPCCA (schur vectors are not orthogonal). I do however feat that this is not very robust. 

## `self_transitions`
The effects of this parameter are not as dramatic as `use_negative_cosines`. Including self-transitions seems to
- help a bit with ill-conditioning
- makes the final states split up earlier into cyclic states
- makes the Delta state appear later
See Fig 3 (no self transitions) and 4 (self transitions) for an example case with `n_states=9`

## `perc`
This eleminates edges in the graph which have low transition probability. Even when I set perc to 15, i.e. the lowest 15 % are eleminated, this still does not do very much, nothing noticeable on the level of the metastable states (it does however, reduces the number of elements in the ranstiion matric, so it works). To check what this really does, one would have to re-run the robustness analysis. 


## Fig 1
image
Recurrent classes when not using negative cosines

## Fig 2
image
Absorption probs towards the recurrent classes of Fig 1

## Fig 3
image
No self-transitions

## Fig 4
image
With self-transitions

Marius1311:
Same bug as in #139, 

```
tests/test_plotting.py:50: in _compare_images
    resize_images_to_same_sizes(expected_path, actual_path)
```

Will go ahead an merge.

Issue #122, opened 2020-05-13
`set_main_states` with no keys/names


Marius1311:
Would be nice if `set_main_states` would just take all metastable states and make them main states in case I call it with `names=None`.  This would often be convenient.

michalk8:
Closed via #128

Issue #117, opened 2020-05-12
Setting main states


michalk8:
Currently, `plot_main_states(n_cells!=None`, ...) writes the `root/final_states` because there
we have the information of how much cells we want to select (no such information is provided in `set_states`).
@Marius1311 how to solve this the best?

Marius1311:
Let's just always write them in `set_main_states`, and let's introduce a parameter for that which is `30` by default. Let's not write them when plotting. This is more expected behaviour, I think. 

So for plotting, the user can select a number of cells, but this will not be written to adata.

michalk8:
Closed via #128

PR #128, opened 2020-05-14
Root final improvements

### LINKED ISSUE #111: Introduce `final_states_cont` and `root_states_cont`
Additionally to the categorical `root_states` and `final_states`, we need continious `root_states_cont` and `final_states_cont` - basically, the probability of being a root/final states, irrespective of _which one_. 

For CFLARE, this already exists, I guess it's currently called `final_cells_probs` and `root_cells_probs`, so here, it's only a matter of renaming. For GPCCA, we can get this simply by adding up our lineage probabilities corresponding to root/final states (not taking into account the `rest` in case we did not redistribute, otherwise it won't work). 

Once this is done, we can nicely interface with scvelo's latent time and directed paga. We're talking to Volker tomorrow, would be nice to have this ready by then so he can start with the interface on scvelo's side.

michalk8:
Fixes #111  #122 , #117 
@Marius1311 I reluctantly did as you've suggested and opted NOT to save to `adata.obs` when plotting `n_cells!=None`. However, I think this might confusing for user, having `X` cells saved and `Y` cells not. The `saving` approach is on the other hand a bit redundant: `plot_main_states` with different that `n_cells` when calling `set_main_states` will be overwritten, making the previous choice useless.
One option would be instead of having `n_cells=...` have `plot_categorical={True,False}`, but this is clunky in a way that user has to call `set_main_states(...)` and then `plot_main_states(...)` to visualize different `#cells`.

### DIFF
--- a/cellrank/tools/_constants.py
+++ b/cellrank/tools/_constants.py
@@ -55,7 +55,7 @@ def _colors(k: Union[str, LinKey, RcKey]) -> str:
 
 
 def _probs(k: Union[str, RcKey]) -> str:
-    return f"{k}_cont"
+    return f"{k}_probs"
 
 
 def _dp(k: Union[str, LinKey]) -> str:
--- a/cellrank/tools/estimators/_gppca.py
+++ b/cellrank/tools/estimators/_gppca.py
@@ -432,8 +432,34 @@ class GPCCA(BaseEstimator):
                 **kwargs,
             )
 
+    def _set_main_states(self, n_cells: Optional[int]) -> None:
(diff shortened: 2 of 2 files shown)


Marius1311:
mhm,  I like the alternative you suggested: setting `n_cells` when calling either `compute_main_states` or `set_main_states`, and changing to `plot_categorical` in the plotting function. That would be very clean in my opinion - we set and write the attributes in the `compute` function and we show them in the `plot` function.

Marius1311:
I don't think it's too clunky.

Issue #124, opened 2020-05-13
Velocity uncertainty interfacing


Marius1311:
scvelo computes variances that we can use to scale the softmax. These variances will have the same shape as the velocity graph. We should settle on a naming and location for this matrix (probably `.uns`), so we can easily read it in.

Only the `VelocityKernel` class will need the uncertainties in the call to `.compute_transition_matrix`.

Marius1311:
This is done for now. We will have to see in practice what the effects of using the velocity uncertainties are.

Marius1311:
Good point, will move this to info level.

PR #129, opened 2020-05-15
Kernel velo vars

michalk8:
Partially implements #124 
I've added `var_key` to the Kernels (not only `VelocityKernel` - I think we shouldn't restrict user much and make it as general as possible). If this is incorrect usage, then we should prevent this and I can remove the code + add some checks that it doesn't happen.
@Marius1311 : am I correct in assuming the following workflow?:
1. compute T
2. scale with variances, if present (just elementwise product) and normalize (x / sum(x))
3. optional density normalization


Marius1311:
Hi Mike, thanks for this! Comments

1. It's okay to allows the `var_key` for every kernel. However, so far, we are only going to take it into account for the velocity kernel.
1. the workflow isn't quite right. The variances are taken into account when we compute T, they are used to scale to softmax. See below

## Transition probabilities computation

So far, we are executing: 

```
# compute directed graph --> multi class log reg
velo_graph = correlations.copy()
velo_graph.data = np.exp(velo_graph.data * sigma_corr)
```

we need to change this to:
```
# compute directed graph --> multi class log reg
velo_graph = correlations.copy()
velo_graph.data = np.exp(velo_graph.data / variances)
```

where we need to make sure that the variances are > 0. Ideally, Volker would save the raw variances and we would clip them to a value `var_min > 0`, like `var_min=0.1`. This should be a parameter of the method.

michalk8:
Of which method - the one that computes the variances or `compute_transition_matrix`? What if variances are not present, are we stil scaling by sigma_corr?

michalk8:
I was wondering what's the best way to store the variances. @VolkerBergen how sparse do you think the matrix will be? If it's going to be mostly sparse, I think we should do the clipping in CellRank.

VolkerBergen:
Has a variance/score for each entry in the `velocity_graph`, thus would be stored in `.obsp` and clipping is sufficient to be done on `sparse_matrix.data` directly.

Marius1311:
> Of which method - the one that computes the variances or `compute_transition_matrix`? What if variances are not present, are we stil scaling by sigma_corr?

The method that computes the variances will be part of scvelo, I am talking about the method that computes the transition probabilities. If variances are not present, we still scale by sigma_corr. 

If we clip variances to `v_min`, i.e. `variance >= v_min` for all entries, then `1/v_min` will the maximum factor we multiply by in the softmax, and this should be in the order of magnitude of `sigma_corr`.

Marius1311:
Any progress on this yet?

PR #130, opened 2020-05-15
Start one state implementation

Marius1311:
This is an initial attempt that I need for the robustness analysis. To complete this, two things are missing form my point of view:

- handling `self._schur_vectors, self._coarse_T, self._coarse_init_dist, self._coarse_stat_dist`. For now, I just set them to None. Running one of the plotting associated plotting functions will however urge me now to `compute_metastable_states`. Either we create dummy data for them (everything will be 1), or we modify the associated error messages to say 'Compute metastable states first using more than 1 state'
- adapt `compute_main_states` to the one-state case: take the one state and make it the main state (no redistribution).


Marius1311:
One more thing: when we use the stationary distribution for `root/final_states_probs`, the probabilities will be much smaller than usually because the stationary distribution is scaled to sum to one. 

Once way to solve this would be to simply scale `root/final_states_probs` such that the largest value is always 1. What's your opinion?

Marius1311:
@michalk8 I'm merging this, should be fine. I adapted the error messages when a user tries to call e.g. `plot_coarse_T` to tell him/her to use `n_states > 1`.

PR #139, opened 2020-05-21
Adapt CFLARE 

Marius1311:
This 
- make the first eigenvector constant for irr. mat, just for plotting (num. errors cause it not to be const. )
- changes naming to be consistent with GPCCA class


Marius1311:
This fails because of 
```
tests/test_plotting.py:50: in _compare_images
    resize_images_to_same_sizes(expected_path, actual_path)
```

Something with these ground truth figures. I don't think this is critical, so I will go ahead and merge this.

Issue #148, opened 2020-06-02
Fix gene trend plotting test


Marius1311:
This is failing because some image is missing. Would be great if we could re-generate the image and fix the test.

michalk8:
There's deeper issue at play - the order of genes is not always preserved (across different scanpy/adata versions?) and I'm using indices. I will hardcode the gene names to prevent this issue from rising up again.

Marius1311:
ok I see

Marius1311:
Ah, so good to have that `build=passing` badge!

PR #150, opened 2020-06-02
Fix passing titles

### LINKED ISSUE #142: Cluster Fates with mode `paga_pie` ignores passed title

michalk8:
Fixes #142


michalk8:
Also fixed #148 - the issue is that the `adata` subset is not always the same (possibly differs form different versions of adata/scanpy - making the var_names unique in https://github.com/theislab/scvelo/blob/master/scvelo/datasets.py#L36).
In our tests, we use indices instead of gene names - changing this would be beneficial.

Issue #138, opened 2020-05-18
Schur decomposition


Marius1311:
Long term, it would be better to compute the schur decomposition _before_ running GPCCA, i.e. having a `self.decompose(mode: str = 'schur/eigen', method: str = 'krylov/brandts/scipy')` method that one calls before calling `self.compute_metastable_states` for the following reaons:

1. efficiency. Often, I want to compute the decomposition for several values of `n_states`. Currently, that means I'm recomputing the schur decomposition every time I do that. 
1. a priori defining a set of feasible values for `n_states`. Currently, I run `compute_metastable_states` and it fails if I split a pair of complex conjugate eigenvalues. Would be nicer to know this a priori. Currently, this is possible by first computing the eigendecomposition, but this is actually a duplication. 
1. being able to inspect the schur vectors before running `compute_metastable_states`
1. would enable us to not have to run `krylov_schur` before method `brandts`. The problem at the moments it: where do I get the eigenvalues from to say whether we will split a block of complex conjugates? I can compute the evals from brandts, but only once I have computed the full decomposition, see #137

Marius1311:
In practical terms, this would mean moving some of the GPCCA code to compute decompositions into cellrank, and creating a possibility in msmtool's GPCCA to use an already existing schur decomposition.

michalk8:
@Marius1311 I don't think we need to move any code, look here: https://github.com/msmdev/msmtool/blob/krylov_schur/msmtools/analysis/dense/gpcca.py#L1109 - it runs the Schur decomposition only when previous one is not found!

Marius1311:
oh, I see! Do we make use of this in our code yet, or do we initialise a new GPCA object every time?

michalk8:
Implemented via #155

PR #155, opened 2020-06-02
Add compute_schur

michalk8:
Implements #138 
@Marius1311 I need your feedback on this. I've decoupled the computation of Schur vectors and the metastable states - there's the function `compute_schur(n_vectors, init_dist, method, which)` which just calls (`_do_schur_helper`). Afterwards, you can plot the vectors/matrix using `plot_schur_embedding()` or `plot_schur_matrix()`.
Then, you can specify number of states in `compute_metastable_states` - if it's <= #previously computed vectors, you don't recompute it, otherwise, it warns you and computes the new decomposition.
In #138 2., you also mention the ED - however, as it currently stands, there's no way for me to access it or pass it to `msmtools.GPCCA`. I'd leave it for now.
Let me know what you think and whether this is what you've wanted. If so, I will merge and start writing tests.

### DIFF
--- a/cellrank/tools/estimators/_gppca.py
+++ b/cellrank/tools/estimators/_gppca.py
@@ -177,7 +177,6 @@ class GPCCA(BaseEstimator):
         self,
         title: Optional[str] = "schur matrix",
         cmap: str = "viridis",
-        upper_triangular_only: bool = True,
         figsize: Optional[Tuple[float, float]] = None,
         dpi: Optional[float] = 80,
         save: Optional[Union[str, Path]] = None,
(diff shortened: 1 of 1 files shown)


Marius1311:
Oh, great, I will go through this later!

Marius1311:
Your proposed workflow above sounds very good to me. Here is what I would do regarding the eigendecomposition:
- msmtools.GPCCA does not compute eigenvectors
- however, in https://github.com/msmdev/msmtools/pull/9, I add the computation of eigenvalues for both method, `brandts` and `krylov` (note that we removed `scipy` as an option, it did not work.) These will be in self.eigenvalues
- so I would suggest that we extract 3 things from msmtools.GPCCA after running `_do_schur_helper`: schur vectors, schur matrix and eigenvalues

Marius1311:
Just pushed some minor changes. One request: for the schur matrix, we can at the moment either compute just the upper half of everything. However, we know by theory that it will be _quasi upper triangular_, meaning that it will be upper triangular except for some 2x2 blocks on the diagonal. Could we change the visualisation to the upper half + any 2x2 blocks that are non-zero?

Marius1311:
image
To make this more obvious, everything below the diagonal is zero, apart from the two 2x2 blocks you can see. These are imortant.

Marius1311:
great, works just as expected!

Issue #163, opened 2020-06-04
High level functions


Marius1311:
We need to add a parameter `mode='GPCCA/CFLARE'` to our high level functions and adapt them for the GPCCA setting.

michalk8:
Implemented via #167

Marius1311:
Awesome, thanks, will check this out later today.

PR #167, opened 2020-06-04
High level fn gpcca

michalk8:
Implements #163 
@Marius1311 I've opted for `estimator` key which is our class, but can change it to strings (have no strong opinions about this).
I wasn't sure how to best set the lineages for GPCCA, so I just use `set_main_states()` without everything.

### DIFF
--- a/cellrank/tools/_lineages.py
+++ b/cellrank/tools/_lineages.py
@@ -103,6 +103,7 @@ def lineages(
 
     # compute the absorption probabilities
     if isinstance(mc, CFLARE):
+        mc.compute_eig()
         mc.compute_metastable_states(**kwargs)
         mc.compute_lin_probs(keys=keys)
     elif isinstance(mc, GPCCA):
--- a/cellrank/tools/_lineages.py
+++ b/cellrank/tools/_lineages.py
@@ -111,10 +111,11 @@ def lineages(
         mc.compute_metastable_states(cluster_key=cluster_key, **kwargs)
         mc.compute_lin_probs(keys=keys)
     elif isinstance(mc, GPCCA):
-        if n_lineages == 1:
+        if n_lineages is None or n_lineages == 1:
             mc.compute_eig()
-        elif n_lineages is not None:
-            mc.compute_schur(n_components=n_lineages + 1, method=method)
+        else:
+            mc.compute_schur(n_lineages + 1)
+
         try:
             mc.compute_metastable_states(
                 n_states=n_lineages, cluster_key=cluster_key, **kwargs
--- a/cellrank/tools/_lineages.py
+++ b/cellrank/tools/_lineages.py
@@ -16,6 +16,7 @@ def lineages(
     adata: AnnData,
     estimator: type(BaseEstimator) = CFLARE,
     final: bool = True,
+    cluster_key: Optional[str] = None,
     keys: Optional[Sequence[str]] = None,
     n_lineages: Optional[int] = None,
     method: str = "krylov",
@@ -50,7 +51,7 @@ def lineages(
         keys=['Neuronal_1, Neuronal_2', 'OPC'] means that the two neuronal endpoints are treated as one and
         Astrocytes are excluded.
     n_lineages
-        Number of lineages when :paramref:`estimator` `=GPCCA`.
+        Number of lineages when :paramref:`estimator` `=GPCCA`. If `None`, it will be based on `eigengap`.
     method
         Method to compute Schur vectors when :paramref:`estimator` is :class:`cellrank.tl.GPCCA`.
     copy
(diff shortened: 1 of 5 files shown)


michalk8:
I've also changed that the high-lvl functions never read fro `adata`, just to let you know (it was confusing me why it was running, because it shouldn't - I have an object where everything is precomputed).

Marius1311:
Ah great, thanks, will have a look at this later!

Marius1311:
After having run `cr.tl.find_final(adata_t, estimator=GPCCA, n_states=3)`, I can't yet plot the `final_states` using `scv.pl.scatter(adata_t, color='final_states')`, that raises `ValueError: color key is invalid! pass valid observation annotation or a gene name`. 

Have you opted for only writing the states in `cr.tl.lineages`? I think this would be a bit upsetting for the user, calling `.find_final` should make it possible to plot the identified states.

Marius1311:
Running `cr.tl.lineages(adata_t, estimator=GPCCA, n_lineages=5)` first gives me a warning: 

```
Computing lineage probabilities towards `final_states`
WARNING: Using 5 components would split a block of complex conjugates. Increasing `n_components` to 6
Computing metastable states
```
But apparently tries with 5 states anyways, which raises an exception then:
```
ValueError                                Traceback (most recent call last)
<ipython-input-27-846879d89ccb> in <module>
----> 1 cr.tl.lineages(adata_t, estimator=GPCCA, n_lineages=5)

~/Projects/cellrank/cellrank/tools/_lineages.py in lineages(adata, estimator, final, keys, n_lineages, method, copy, return_estimator, **kwargs)
    118         else:
    119             mc.compute_schur(n_components=n_lineages, method=method)
--> 120         mc.compute_metastable_states(n_states=n_lineages, **kwargs)
    121         mc.set_main_states(names=keys)
    122     else:

~/Projects/cellrank/cellrank/tools/estimators/_gppca.py in compute_metastable_states(self, n_states, use_min_chi, n_cells, cluster_key, en_cutoff, p_thresh)
    426             start = logg.info("Computing metastable states")
    427 
--> 428             self._gpcca = self._gpcca.optimize(m=n_states)
    429 
    430             # when `n_cells != None` and the overlap is high, we're skipping some metastable states

~/Projects/msmtools/msmtools/analysis/dense/gpcca.py in optimize(self, m, return_extra)
   1320 
   1321         # Calculate Schur matrix R and Schur vector matrix X, if not adequately given.
-> 1322         self._do_schur_helper(max(m_list))
   1323 
   1324         # Initialize lists to collect results.

~/Projects/msmtools/msmtools/analysis/dense/gpcca.py in _do_schur_helper(self, m)
   1119                     else:
   1120                         if _check_conj_splitting(self.eigenvalues, m):
-> 1121                             raise ValueError(f'Clustering into {m} clusters will split conjugate eigenvalues. '
   1122                                              f'Request one cluster more or less. ')
   1123                         print('INFO: Using pre-computed schur decomposition')

ValueError: Clustering into 5 clusters will split conjugate eigenvalues. Request one cluster more or less. 
```
Not quite sure why this is happening.

michalk8:
Not sure what's happening there either, will look into it. Once #170 passes, I add the option with eigengap + `cluster_key` argument.

michalk8:
@Marius1311 all the issues should be addressed by now, I'd be glad if you could give it a go. Question: there seems to be no conceptual difference for `find_root/final` and `lineages` for GPCCA. Do we want to do anything about it?

Marius1311:
Okay, I'm on it

Marius1311:
I would make the following change: if `cluster_key` is None, then by default, it should be set to either 'clusters' or 'louvain', if any of these is present in `adata.obs`. (If both are present, let's use 'clusters')

Marius1311:
By the way, you are right regarding there being no conceptual differences between `_root_final` and `lineages`, (apart from the `keys`) but I think that's fine for now.

michalk8:
I must advise against this - we can throw a warning that such key(s) are present, but I wouldn't set it.
It's because sometimes the user may want to it to be associated with clusters (we'd need to introduce another argument to handle it).

Marius1311:
Okay, then let's just print out a warning that it's present.

Marius1311:
I just realised one thing, I think the eigengap computation does not work yet, calling 
`cr.tl.find_final(adata_t, estimator=GPCCA, cluster_key='clusters')`
raises
```
RuntimeError                              Traceback (most recent call last)
<ipython-input-46-54b28fe9d170> in <module>
      1 from cellrank.tools import GPCCA
      2 
----> 3 cr.tl.find_final(adata_t, estimator=GPCCA, cluster_key='clusters')
      4 scv.pl.scatter(adata_t, color='final_states')

~/Projects/cellrank/cellrank/tools/_root_final.py in find_final(adata, estimator, n_states, weight_connectivities, show_plots, copy, return_estimator, **kwargs)
    187         copy=copy,
    188         return_estimator=return_estimator,
--> 189         **kwargs,
    190     )

~/Projects/cellrank/cellrank/tools/_root_final.py in _root_final(adata, estimator, final, n_states, cluster_key, weight_connectivities, show_plots, copy, return_estimator, **kwargs)
     99         try:
    100             mc.compute_metastable_states(
--> 101                 n_states=n_states, cluster_key=cluster_key, **kwargs
    102             )
    103         except ValueError:

~/Projects/cellrank/cellrank/tools/estimators/_gppca.py in compute_metastable_states(self, n_states, use_min_chi, n_cells, cluster_key, en_cutoff, p_thresh)
    360             if self.eigendecomposition is None:
    361                 raise RuntimeError(
--> 362                     "Compute eigendecomposition first as `.compute_eig()` or `.compute_schur()`."
    363                 )
    364             was_from_eigengap = True

RuntimeError: Compute eigendecomposition first as `.compute_eig()` or `.compute_schur()`.
```

Have you already added the computation of the eigengap to `.compute_schur`?

michalk8:
Yes, I've added it, but there's a faulty logic there - thanks for spotting it out, I will fix it.
We really need to start writing tests now.

Marius1311:
I was just thinking the same....

Issue #158, opened 2020-06-03
Move gdpt to GPCCA estimator


Marius1311:
GDPT works with the schur decomposition, which is not available in CFLARE. Therefore, the method should be moved. Also, as soon as we have the schur decomposition available, it should use it, rather than recomputing it.

michalk8:
Implemented via #171

PR #171, opened 2020-06-04
Move GDPT

michalk8:
Implements #158


Marius1311:
Nice! I just had a quick look and it seems like this already makes use of the pre-computed schur decomposition? This would be very nice.

michalk8:
Yes, it's already using precomputed Schur vectors. Can I merge it?

Marius1311:
Sure, go ahead.

Marius1311:
Do we generally make sure we fall back to `brandts` when to SLEPSc is installed, whenever `.compute_schur` is called? This would be important.

michalk8:
That behavior is implemented within `msmtools`.

PR #172, opened 2020-06-05
Switch the default estimator to be GPCCA

Marius1311:
Sorry, that commit message below is confusing. I'm really switiching the default estimator to be GPCCA.


Marius1311:
I'm getting rid of using the words `start points` and `endpoints`, as well as `final_cells` and `root_cells` everywhere and replacing this by `final_states` and `root_states` every where to be consistent.

PR #183, opened 2020-06-08
Possible indexing fix

### LINKED ISSUE #181: Bug in `self._select_cells`.
I compute metastable states as `g_fwd.compute_metastable_states(cluster_key='clusters', n_states=2)`, which gives the following output:
```
Computing metastable states
INFO: Using pre-computed schur decomposition
WARNING: The following groups could not be mapped uniquely: `AT2 cells_1, AT2 cells_2`
Adding `.schur_vectors`
       `.metastable_states`
       `.coarse_T`
       `.coarse_stationary_distribution`
    Finish (0:00:00)
```
The resulting states look like this: 
image
This makes no sense. So I check the membership vectors themselves:
image

They look fine, just the names are off, see the cluster assignment below:
image
This tells me: the first state should be called "to ciliated cells", while the second one should be called "to AT2 cells" 

Also, I check the metastable assignment like this: `scv.pl.scatter(adata, color=g_fwd._gpcca.metastable_assignment, color_map='Paired')`, which gives
image

This also looks fine, so it must be a problem with selecting cells from each metastable state. It seems like the wrong cells are selected, which subsequently means that the state gets the wrong name and color assigned. 

If needed, I can send you the anndata object to reproduce. Have we got tests yet for 
- selecting cells from each metastable state
- giving names to metastable states
- assigning colors to metastable states
? 

Since these seem to be a bit tricky in practice, would be important to include some tests for them.

michalk8:
Fixes #181 
@Marius1311 could you please test this out where it fixes the problem? If yes, I will begin writing tests.

### DIFF
--- a/cellrank/tools/estimators/_gppca.py
+++ b/cellrank/tools/estimators/_gppca.py
@@ -945,8 +945,12 @@ class GPCCA(BaseEstimator):
             else metastable_states.droplevel(0)
         )
 
-        metastable_states = metastable_states.sort_index().astype("category")
-        metastable_states.index = self.adata.obs.index
+        metastable_states = metastable_states.astype("category")
+        metastable_states = metastable_states[self.adata.obs_names]
+
+        assert np.all(
+            metastable_states.index == self.adata.obs_names
+        ), "Index mismatch."
 
         # postpone warning after we optionally match the names using `cluster_key`
         if warn:
--- a/cellrank/tools/estimators/_gppca.py
+++ b/cellrank/tools/estimators/_gppca.py
@@ -883,7 +883,11 @@ class GPCCA(BaseEstimator):
         warn: bool = True,
     ) -> Tuple[pd.Series, Union[List[str], List[int]]]:
         def set_categories(group):
+            if not len(group):
+                return group
+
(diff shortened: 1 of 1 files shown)


Marius1311:
Thanks, will try this.

Marius1311:
This works, great!

michalk8:
Ok, merging this, I've also fixed a possible issue with having an empty category. I didn't add any tests yet because in the current implementation, it's not that easy. I think it might be worthwhile to refactor some of these to utils (i.e. be class independent) so that it's easier to test.

cellrank/tools/estimators/_gppca.py
```diff
@@ -933,15 +938,19 @@ def set_categories(group):
         meta_assignment["assignment"] = (
             meta_assignment["assignment"].astype(str).astype("category")
         )
+        meta_assignment["assignment"].cat.set_categories(["0", "1", "2"], inplace=True)
```
Marius1311, line 941: Why do you hardcode this here?

michalk8: This was when I was testing this and I forgot it there, sorry.

PR #184, opened 2020-06-09
Refactor select cells

Marius1311:
This changes the way in which we select cells. The central piece is a new utility function `_fuzzy_to_discrete`. This function is class-independent and should make it easier to write tests. 

The new selection function considers overlaps between all states and offers two different possibilities for dealing with it, see the docstring of `_fuzzy_to_discrete` for more details.

### DIFF
--- a/cellrank/tools/_utils.py
+++ b/cellrank/tools/_utils.py
@@ -1232,9 +1232,6 @@ def _fuzzy_to_discrete(
     if raise_threshold is not None:
         n_raise = int(raise_threshold * n_most_likely)
 
-    # create argmax assignment as a lookup table
-    max_assignment = a_fuzzy.argmax(axis=1)
-
     # initially select `n_most_likely` samples per cluster
     sample_assignment = {
         cl: fuzzy_assignment.argpartition(-n_most_likely)[-n_most_likely:]
@@ -1252,7 +1249,11 @@ def _fuzzy_to_discrete(
         if remove_overlap:
             a_discrete[sample_ix, :] = _one_hot(n_clusters)
         else:
-            a_discrete[sample_ix, :] = _one_hot(n_clusters, max_assignment[sample_ix])
+            candidate_ixs = np.where(a_discrete[sample_ix, :])[0]
+            most_likely_ix = candidate_ixs[
+                np.argmax(a_fuzzy[sample_ix, a_discrete[sample_ix, :]])
+            ]
+            a_discrete[sample_ix, :] = _one_hot(n_clusters, most_likely_ix)
 
     # check how many samples this left for each cluster
     n_samples_per_cluster = a_discrete.sum(0)
@@ -1262,6 +1263,8 @@ def _fuzzy_to_discrete(
                 f"Discretizing leads to a cluster with less than {n_raise} samples. "
                 f"Consider recomputing the fuzzy clustering."
             )
+    if (n_samples_per_cluster > n_most_likely).any():
+        raise ValueError("Assigned more samples than requested. ")
     critical_clusters = np.where(n_samples_per_cluster < n_most_likely)[0]
 
     return a_discrete, critical_clusters
--- a/cellrank/tools/_utils.py
+++ b/cellrank/tools/_utils.py
@@ -17,12 +17,6 @@ from typing import (
 )
 from itertools import tee, product, combinations
 
-import matplotlib.colors as mcolors
-
-import scanpy as sc
-from scanpy import logging as logg
-from anndata import AnnData
-
 import numpy as np
 import pandas as pd
 import networkx as nx
@@ -33,6 +27,13 @@ from sklearn.cluster import KMeans
 from pandas.api.types import infer_dtype, is_categorical_dtype
 from sklearn.neighbors import NearestNeighbors
 from scipy.sparse.linalg import norm as s_norm
+
+import matplotlib.colors as mcolors
+
+import scanpy as sc
+from scanpy import logging as logg
+from anndata import AnnData
+
 from cellrank.utils._utils import _get_neighs, _has_neighs, _get_neighs_params
 from cellrank.tools._colors import _convert_to_hex_colors, _insert_categorical_colors
 
@@ -1158,12 +1159,10 @@ def _one_hot(n, cat: Optional[int] = None) -> np.ndarray:
 
     If cat is None, return a vector of zeros.
     """
-    if cat is not None and cat >= n:
-        raise ValueError(f"Can't one-hot encode {cat} with a vector of length only {n}")
 
-    out = np.zeros(n)
+    out = np.zeros(n, dtype="bool")
     if cat is not None:
-        out[cat] = 1
+        out[cat] = True
 
     return out
 
@@ -1237,9 +1236,9 @@ def _fuzzy_to_discrete(
     }
 
     # create the one-hot encoded discrete clustering
-    a_discrete = np.zeros_like(a_fuzzy)
+    a_discrete = np.zeros_like(a_fuzzy, dtype="bool")
     for ix in range(n_clusters):
-        a_discrete[sample_assignment[ix], ix] = 1
+        a_discrete[sample_assignment[ix], ix] = True
 
     # handle samples assigned to more than one cluster
     critical_samples = np.where(a_discrete.sum(1) > 1)[0]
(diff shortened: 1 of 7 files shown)


Marius1311:
Also, I would vote for removing `n_cells` from all plotting functions. It's confusing to be able to set this parameter through different calls. I would replace it with a parameter `discretize: bool = True`. If `True`, we show the cells selected from each metastable state during `self.compute_metastable_states`. If `False`, we mimic the behaviour of `n_cells = None`. `same_plot` is unaffected by this.

cellrank/tools/_utils.py
```diff
@@ -1150,3 +1150,156 @@ def _info_if_obs_keys_categorical_present(
...
+        index = range(n_samples)
+    if names is not None:
+        if len(names) != n_clusters:
+            raise ValueError(
+                f"Shape mismatch, length of `names` is `{len(names)}`, but `n_clusters` = {n_clusters}."
+            )
+    else:
+        names = np.arange(n_clusters).astype("str")
+
+    target_series = pd.Series(index=index, dtype="category")
```
michalk8, line 1300: How efficient is this?

cellrank/tools/_utils.py
```diff
@@ -1150,3 +1150,156 @@ def _info_if_obs_keys_categorical_present(
...
+    max_assignment = a_fuzzy.argmax(axis=1)
+
+    # initially select `n_most_likely` samples per cluster
+    sample_assignment = {
+        cl: fuzzy_assignment.argsort()[::-1][:n_most_likely]
+        for cl, fuzzy_assignment in enumerate(a_fuzzy.T)
+    }
+
+    # create the one-hot encoded discrete clustering
+    a_discrete = np.zeros_like(a_fuzzy)
```
michalk8, line 1240: Should be the dtype something like `int32`?

cellrank/tools/_utils.py
```diff
@@ -1150,3 +1150,156 @@ def _info_if_obs_keys_categorical_present(
...
+    n_clusters = a_fuzzy.shape[1]
+    if raise_threshold is not None:
+        n_raise = int(raise_threshold * n_most_likely)
+
+    # create argmax assignment as a lookup table
+    max_assignment = a_fuzzy.argmax(axis=1)
+
+    # initially select `n_most_likely` samples per cluster
+    sample_assignment = {
+        cl: fuzzy_assignment.argsort()[::-1][:n_most_likely]
```
michalk8, line 1235: I think it'd be better to use argpartition (it's O(n)) vs argsort (O(n log n)).

cellrank/tools/_utils.py
```diff
@@ -1150,3 +1150,156 @@ def _info_if_obs_keys_categorical_present(
...
+def _one_hot(n, cat: Optional[int] = None) -> np.ndarray:
+    """
+    One-hot encode cat to a vector of length n.
+
+    If cat is None, return a vector of zeros.
+    """
+    if cat is not None and cat >= n:
+        raise ValueError(f"Can't one-hot encode {cat} with a vector of length only {n}")
+
+    out = np.zeros(n)
```
michalk8, line 1164: Dtype missing.

cellrank/tools/_utils.py
```diff
@@ -1150,3 +1150,156 @@ def _info_if_obs_keys_categorical_present(
...
             break
+
+
+def _one_hot(n, cat: Optional[int] = None) -> np.ndarray:
+    """
+    One-hot encode cat to a vector of length n.
+
+    If cat is None, return a vector of zeros.
+    """
+    if cat is not None and cat >= n:
```
michalk8, line 1161: I don't think this check is necessary - it just prettifies the `IndexError` we'd get if it wasn't there.

cellrank/tools/_utils.py
Marius1311: I don't think it's really going to matter since the number of states is always very low.

cellrank/tools/_utils.py
Marius1311: good idea. This matrix will only store 0s and 1s, so could even make it a bool type, no?

cellrank/tools/_utils.py
Marius1311: oh, does argpartition do the same thing?

cellrank/tools/_utils.py
Marius1311: should we make this bool as well?

cellrank/tools/_utils.py
Marius1311: right, will remove this.

cellrank/tools/_utils.py
Marius1311: changed to bool

cellrank/tools/_utils.py
Marius1311: changed to bool

cellrank/tools/_utils.py
Marius1311: okay, looked it up and got it. Will change this to using argpartition.

Marius1311: done.

michalk8:
One more issue: using extremely large cells `n_cells=5000` will cause the argpartition to fail (e.g. for pancreas `5000`), so there needs to be a check. Next issue I had is this:
```python
raise ValueError(
                f"Discretizing leads to a cluster with less than {n_raise} samples. "
                f"Consider recomputing the fuzzy clustering."
            )
```
If I select such large number of cells, it should select all available, but it raises this exception instead (I tried e.g. `5000`).

michalk8:
Yet another issue - this doesn't seem very robust. Here I select `1000`, `1200` and `1500` cells from the pancreas data, 3 MS states:

michalk8:
Whereas previously I've gotten (also with a warning that insufficient #cells were found for `Epsilon`, `Alpha, Epsilon` and `Alpha, Beta, Epsilon` for `1000`, `1200` and `1500` cases):

Marius1311:
I see your point. However, this is a very unrealistic test case and I added a check for `n_cells`. Basically, this function is meant to select a few cells from each metastable state that can be used to characterize this state, i.e. to map a name and color to it. 

In the cases you show above, it's intuitively not possible to uniquely give them a name - they cover a lot of different clusters!

Marius1311:
You did however, point me to a related bug in the code....

Marius1311:
I'm going to bed now - we can continue discussing this tomorrow. I fixed one bug with the assignment - your above examples should look different now (it won't let you select quite as many cells, the cutoff is n_cells/n_clusters, so ~700 in this dataset.

Marius1311:
Oh, and if you want to have cells which are assigned to several states removed, simply set `REMOVE_OVERLAP = True`.

Marius1311:
The tests will fail, we will have to adapt the number of cells we use for testing, but that's okay, the original number was a bit unreasonable anyways. We have to keep in mind the application this is designed for. Good night!

Marius1311:
Hi Mike, I would need your help with two things:
1. Could you have a look over my tests? This is the first time I'm doing this, did I get it right/what can I improve for next time?
1. There is one failing test that I created. It fails because `_fuzzy_to_discrete` can't really handle Lineage objects. Best would be to check in the function whether a Lineage object has been passed, and just use `.X` in such a case, but I did not know how to perform such a check because of circular imports. Could you fix this please? You know the lineage class better than I do. Thanks.

Marius1311:
Thanks Mike!!

cellrank/tools/_utils.py
```diff
@@ -1293,11 +1296,19 @@ def _series_from_one_hot_matrix(
         Pandas Series, indicating cluster membership for each sample. The dtype of the categories is `str`
         and samples that belong to no cluster are assigned `NaN`.
     """
-    n_clusters, n_samples = a.shape[1], a.shape[0]
-    if not type(a) == np.ndarray:
+    n_samples, n_clusters = a.shape
+    if not isinstance(a, np.ndarray):
+        raise TypeError(
+            f"Expected `a` to be of type `numpy.ndarray`, found `{type(a).__name__!r}`."
```
Marius1311, line 1302: Note to self: make sure the matrix is an array in `_series_from_one_hot_matrix` as well.

tests/test_lineage.py
```diff
@@ -240,6 +240,29 @@ def test_row_subset_with_ints(self):
 
         np.testing.assert_array_equal(x[[1, 2, 3], :], np.array(y))
 
+    def test_column_subset_boolean(self):
```
Marius1311, line 243: Nice that you added tests for these.

PR #200, opened 2020-06-22
Add high level function to compute lineage drivers

Marius1311:
So far, this is only possible via the low-level interface. Add possiblitiy to call this from the high level interface.

### DIFF
--- a/cellrank/tools/__init__.py
+++ b/cellrank/tools/__init__.py
@@ -6,5 +6,5 @@ from cellrank.tools._lineages import lineages
 from cellrank.tools.estimators import GPCCA, CFLARE
 from cellrank.tools._root_final import root_states, final_states
 from cellrank.tools._exact_mc_test import exact_mc_perm_test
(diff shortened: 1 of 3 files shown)


Marius1311:
Seems to work, but needs at least very basic tests to make sure works fine for forward/backwards cases.

Marius1311:
@michalk8 , this is a bit clunky because I have a lot of repeated documentation - is there a way to make this more elegent? In essence, I would like to re-use part of the docstring of the low-level method `self.compute_lineage_drivers`.

michalk8:
We could try: https://github.com/Chilipp/docrep

Marius1311:
This looks pretty cool, I just didn't figure our from their documentation whether this would work if the functions are in different python files.

Marius1311:
I will attempt to merge this later today without docrep, but I will open a low-priority issue so we can do this later.

Marius1311:
I added tests for the high level pipelines and check whether all keys are created, should be fine now.

PR #201, opened 2020-06-22
Fix Lineage bug when names/rows are of full shape

### LINKED ISSUE #198: Lineage object does not preserve column order
Under very particular circumstances, the Lineage class does not preserve the correct order of columns. Say I have a lineage object  `L` with 3 columns and names `['Alpha', 'Beta', 'Epsilon']`. If I now subset the rows and columns at the same time, and give all three keys but in a different order, i.e.

```
# create some bool mask
mask = adata.obs['clusters'] == 'Alpha'
mask = list(np.array(mask, dtype=np.bool))

# slice
L[mask, ['Epsilon', 'Alpha', 'Beta']]
```
The returned object will **not** have columns in the order `['Epsilon', 'Alpha', 'Beta']`. This leads to bugs in the `cluster_fates` function when selecting all lineages, but in a different order. In my case, that's important for plotting as I want the order to be consistent across methods.

michalk8:
Fixes #198

### DIFF
--- a/cellrank/tools/_lineage.py
+++ b/cellrank/tools/_lineage.py
@@ -223,7 +223,11 @@ class Lineage(np.ndarray):
                 if item_0.dtype != np.bool:
                     if not issubclass(item_0.dtype.type, np.integer):
                         raise TypeError(f"Invalid type `{item_0.dtype.type}`.")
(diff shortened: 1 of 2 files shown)


Marius1311:
Happy to report that this works!

PR #211, opened 2020-06-25
Update the installation instr.

Marius1311:
A collection of problems encountered at installation at their solution. Related to SLEPSc/PETSc/MPI


Marius1311:
Neet to add further stuff - can install quite a few things from conda.

Issue #214, opened 2020-06-26
Implement kernel that just takes in pre-computed transition matrix


Marius1311:
This would be super useful when I have computed a transition matrix in whatever way and would like to process it using cellrank. Currently, this is quite tricky - I always need to create a dummy adata object etc.

michalk8:
Implemented via #215

PR #215, opened 2020-06-26
Precomputed kernel

michalk8:
Implements #214 
It will create an empty adata object if None is specified, because we save stuff there.
You can still include it if needed.


Marius1311:
wonderful, thanks, I will check this out!

michalk8:
All the tests have passed, so I've merged it.

Marius1311:
Fine, just tested, works well.

PR #216, opened 2020-06-26
Fix one error with kmeans clusterin

Marius1311:
There was a bug in Kmeans clustering when the number of samples was smaller than the number of clusters.


Marius1311:
Okay, I have to fix this up.

Marius1311:
This should fix the problem. It also adds some tests for the underlying utility function, which I also cleaned up a bit.

Issue #204, opened 2020-06-23
Make sure plotting functions don't select cells


Marius1311:
In practice, it just gets really confusing with several calls writing metastable states, ie `self.compute_metastable_states` and `self.plot_metastable_states`. Therefore, I would change this such that

- only `self.compute_metastable_states` writes metastable states, ie selects cells and maps these towards precomputed clusters to get names and colors
- plotting functions just plot however many cells have previously been assigned
- plotting functions will no longer have `n_cells` but `discrete: bool = True`.

Marius1311:
Equally, `set_main_states` should not do the cell selection. This should literally only be done once - when we compute the metastable states.

michalk8:
`{set, compute}_main_states` must select cells, because there we merge the states + redistribute the probability mass.
I agree though that from the plotting functions should not select cells.

Marius1311:
agreed.

michalk8:
Implemented via #217 .

PR #217, opened 2020-06-27
Remove cell selection plotitng

michalk8:
Implements https://github.com/theislab/cellrank/issues/204


michalk8:
There's the new option called `discrete=False` (instead of `n_cells=None`.

PR #218, opened 2020-06-28
Speedup for CFLARE

Marius1311:
This fixes a stupid bottleneck in the computation of abs probs which was related to selecting indices. More importantly, it enables using iterative solvers for the linear problem. So far, I'm using GMRES, but bicgstag also looked promising. The iterative solvers are powerful if the matrices are sparse, very large and I'm not computing too many absorbing cells.

### DIFF
--- a/cellrank/tools/estimators/_cflare.py
+++ b/cellrank/tools/estimators/_cflare.py
@@ -9,7 +9,7 @@ from pandas import Series
 from scipy.stats import zscore, entropy
 from scipy.linalg import solve as solve
 from scipy.sparse import issparse
-from scipy.sparse.linalg import gmres, spsolve
+from scipy.sparse.linalg import gmres
 
 import matplotlib as mpl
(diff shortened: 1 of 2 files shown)


Marius1311:
On 32k cells and 197 absorbing states, I get
- iterative solver: 195s, basically all of them spent in computing matrix vector products. 
- direct solver: 260s, all of which spend in the solve routine. 

The resulting lineage probabilities look visually exactly the same. I haven't looked into them quantitatively yet. Should add a test that checks for similarity, using both solvers.

Note: these settings are actually not very favourable for the iterative solver as it's way to many absorbing states.

Marius1311:
In a favourable setting with only 12 absorbing states, the iterative solver takes 28s, demonstrating it's clear superiority.

Marius1311:
to do's:
- check whether absorption probs agree quanitativly
- add a check for this

PR #223, opened 2020-07-01
Further speedup for CFLARE

Marius1311:
Uses the solutions of old problems to initialize the new one, only within one recurrent class.

### DIFF
--- a/cellrank/tools/estimators/_cflare.py
+++ b/cellrank/tools/estimators/_cflare.py
@@ -657,8 +657,10 @@ class CFLARE(BaseEstimator):
 
         # create arrays of all recurrent and transient indices
         mask = np.repeat(False, len(metastable_states_))
+        rec_start_indices = [0]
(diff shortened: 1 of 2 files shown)


Marius1311:
Currently reduces the compute time on 32k cells for 482 absorbing states from 6:07 min to 5:55 min.

Marius1311:
On just 12 absorbing states, it increases the compute time by 4s from 28 to 32s

Marius1311:
On 69 absorbing states, using the initialisation starts to become an advantage:
with initialisation: 1:15 min
without initialisation: 1:23
with direct solver: 4:29 min

Just need to write one test for this.

Marius1311:
Added a test for this. That was defenitely enough speedup for now. In the future, can experiment with scipy's other iterative solvers optimized for sparse problems. However, if only using few final cells, the problem is solved as we are essentially linear in cell number now.

PR #144, opened 2020-05-26
Scanpy 1.5 compatibility

Marius1311:
AnnData >= 0.7 stores (n_obs x n_obs) matrices in `adata.obsp`. We should adapt our graph data storage accordingly, while maintaining backwards compatibility. This adds some first utility functions.

Issue #141, opened 2020-05-25
Scanpy 1.5 compatibility


Marius1311:
AnnData >= 0.7 stores graph related data in `adata.obsp`, we should take this into account when reading/writing attributes, while keeping backward compatibility

Marius1311:
Simple need a function which first looks in `.obsp` for the attribute and, if it's not there, looks in `.uns` and issues a warning.

Marius1311:
Started this in #144

michalk8:
So, is this an `scanpy` versioning problem or `anndata` or both? What I'd do is to modify the functions you've created and from the mentioned version(s), default to `obsp` and fallback to `uns` or vice versa.
Question is, how compatible do you want it to be (asking about whether we should default to anything at all)? Cleanest way would be to look only in `obsp` from >= 0.7 and fail if there's nothing (let the user move it). The only benefit for defaulting to something different (`uns` or `obsp`) is that user doesn't have to move the data themself.

Marius1311:
I think it's both `scanpy` and `anndata`: anndata created the new `.obsp` field and scanpy started storing the output of `.pp.neighbors` in `.obsp` from version 1.5, see the changelog [here](https://scanpy.readthedocs.io/en/latest/)

So let me make sure I understand what you are saying: if the anndata version is >= 0.7, we check in `obsp` and we don't default to `.uns`. If the anndata version is <0.7, we just check in `.uns`?

@VolkerBergen , how are you going to handle this in scvelo?

VolkerBergen:
I created a helper function to maintain compatibility with stored data from older `anndata` versions.
```
def get_neighs(adata, mode="distances"):
    return (
        adata.obsp[mode]
        if hasattr(adata, "obsp") and mode in adata.obsp.keys()
        else adata.uns["neighbors"][mode]
    )
```
In scvelo, `distances` and `connectivities` are stored in `obsp` along the same lines as in scanpy, while the `velocity_graph` and all other pairwise annotations will be moved to `obsp` as well in the next release.

As a side note, there's currently some memory issues with `obsp`, so it will only be fully functional with the next anndata release.

michalk8:
@Marius1311 Is `.obsp` only for square-matrix data, like `cell-to-cell` or `gene-to-gene` graphs? If not, I'd store `Lineage` there as well.

michalk8:
Implemented via #226

Marius1311:
[...]

Issue #94, opened 2020-04-29
PyPI and Anaconda


michalk8:
Start slowly preparing to do this.

michalk8:
Now that #178 is merged, I will focus on this.

michalk8:
Hi, @flying-sheep,
I'm struggling to put CellRank on `testpypi`, the main culprit being here: https://github.com/theislab/cellrank/blob/master/setup.py#L42
The problem is this: `HTTPError: 400 Client Error: Invalid value for requires_dist. Error: Can't have direct dependency: 'msmtools @ git+git://github.com/msmdev/msmtools@krylov_schur#egg=msmtools' for url: https://test.pypi.org/legacy/`
This problem occures when I build `bdist_wheel`. If I only build `sdist`, it gets uploaded.
However, after running `pip install -i https://test.pypi.org/simple/ cellrank==1.2.3.dev2`,
I get the following error and I don't know how to fix it:
```python
Installing build dependencies ... error
  ERROR: Command errored out with exit status 1:
   command: /home/michal/.miniconda3/envs/cellrank/bin/python /home/michal/.miniconda3/envs/cellrank/lib/python3.8/site-packages/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-ant09w4q/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://test.pypi.org/simple/ -- setuptools setuptools_scm wheel
       cwd: None
  Complete output (3 lines):
  Looking in indexes: https://test.pypi.org/simple/
  ERROR: Could not find a version that satisfies the requirement setuptools (from versions: none)
  ERROR: No matching distribution found for setuptools
  ----------------------------------------
ERROR: Command errored out with exit status 1: /home/michal/.miniconda3/envs/cellrank/bin/python /home/michal/.miniconda3/envs/cellrank/lib/python3.8/site-packages/pip install --ignore-installed --no-user --prefix /tmp/pip-build-env-ant09w4q/overlay --no-warn-script-location --no-binary :none: --only-binary :none: -i https://test.pypi.org/simple/ -- setuptools setuptools_scm wheel Check the logs for full command output.
```
Here's the full [log](https://github.com/theislab/cellrank/files/4745534/log.txt).

Based on [this](https://stackoverflow.com/questions/54887301/how-can-i-use-git-repos-as-dependencies-for-my-pypi-package#answer-54894359), I'm not sure if it's even possible to specify dependency that way.
[...]

Issue #222, opened 2020-06-30
Further speed up CFLARE by providing initial guess for gmres


Marius1311:
The sequence of linear problems will be very similar. Exploit this for starting guesses.

Marius1311:
closed by #223

PR #225, opened 2020-07-01
Fix versioning

michalk8:
Implements #94

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.4.0.a0
+current_version = 1.2.0
 commit = True
 tag = True
 files = setup.py cellrank/__init__.py
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,11 +1,11 @@
 [bumpversion]
-current_version = 1.2.1.a0
+current_version = 1.0.0-a.0
 commit = True
 tag = True
 files = setup.py cellrank/__init__.py
-parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\.(?P<release>[a-z]+)(?P<n>\d+))?
+parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(-(?P<release>[a-z]+)\.(?P<n>\d+))?
 serialize = 
-	{major}.{minor}.{patch}.{release}{n}
+	{major}.{minor}.{patch}-{release}.{n}
 	{major}.{minor}.{patch}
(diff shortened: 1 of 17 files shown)


michalk8:
@Marius1311 - I think we will be able to put this on PyPI soon. However I'd like these to be merge before an official release:
- [ ] https://github.com/theislab/cellrank/issues/222 (Marius) - is this straightforward?
- [ ] https://github.com/theislab/cellrank/issues/219 (Marius) - I think this can be done pretty fast
- [ ] https://github.com/theislab/cellrank/issues/141 (Michal) - I suggest requiring `scanpy>=1.5` (scvelo does it as well) and save the transition matrix in `.obsp`
- [ ] https://github.com/theislab/cellrank/issues/192 (Michal) - the problem seems to be with the bashuploader - I've changed the workflow, not using the `codecov-actions`, but only the uploader, and it still doesn't work (because the uploader detects whether we're in an Action or not) - I suggest doing it manually for now, until someone responds on `codecov-actions` with a solution.

Also we should start documenting the releases in our docs.

michalk8:
@Marius1311 Can you try downloading from PyPI test?
`pip install -i https://test.pypi.org/simple/ cellrank==1.2.1a0`

Marius1311:
Thanks Mike! #222 is done. #219 mainly aims at our transition matrix to be the same as scanpy's in case we only use a `ConnectivityKernel`. I will open a PR for this with an associated test that makes this more obvious. I should be able to do this later today or tomorrow.

Marius1311:
Re #141, fine by me to require scanpy>=1.5. Feel free to make the required changes and to close #144

Marius1311:
Re #192, maybe Phil can help? manual would be fine with me for now as well.

Marius1311:
Okay, running `pip install -i https://test.pypi.org/simple/ cellrank==1.2.1a0` gives me
```
ERROR: Could not find a version that satisfies the requirement matplotlib>=3.0.3 (from cellrank==1.2.1a0) (from versions: 2.2.3)
ERROR: No matching distribution found for matplotlib>=3.0.3 (from cellrank==1.2.1a0)
```

Marius1311:
I'm running python 3.6 in a new conda environment if that helps.

Marius1311:
#219 is not super relevant at the moment, we can safely do this after pypi release.

michalk8:
Ok, let's do it as follows: I will fix #141 and post in on PyPI as pre-release (i.e. `v1.0.0-alpha` or similar). Once the manuscript is done (and other issues are fixed), we'll bump the version.

Marius1311:
sounds good to me.

PR #231, opened 2020-07-03
Move absorption probabilities to the base estimator

Marius1311:
Absorption probabilities can be computed, no matter how the metastable states have been found. Therefore, I would treat it as a general concept and move it to the base estimator.


Marius1311:
The idea is to move general ideas to the base estimator, so that both CFLRE and GPCCA can compute absorption probs. This has become relevant now that I managed to scale up the computation of absorption probabilities.

Marius1311:
This is not the ideal solution yet as it's a bit unclear how to properly integrate this with GPCCA. However, it works and it does not break anything else, so I'm merging. This makes it easier for me to compare absorption probabilities with GPCCA membership vectors.

Issue #236, opened 2020-07-03
Change the default colormap to be colorblind-friendly


Marius1311:
So far, when I don't provide a cluster key, I get a fairly standard matplotlib color palette:
![image](https://user-images.githubusercontent.com/24494650/86479740-06a79980-bd4d-11ea-9a1a-7fcaaa2af6a5.png)

However, using red and green together is really not very colorblind friendly. Volker takes care of it [here](https://github.com/theislab/scvelo/blob/a7de78ae974b251c366f208aab2e685b23bc101e/scvelo/settings.py#L80) in scvelo. Would be nice to just use the same colormap by default for categorical annotations.

michalk8:
Ok, I've changed the order to `tab10`, `tab20`, etc. This should've fixed it.

Marius1311:
fab, will try!

PR #237, opened 2020-07-03
Fix palettes

michalk8:
Implements #236


Marius1311:
I added `Dark2`, I have a feeling this may have messed up some tests...

Marius1311:
I reverted that commit, maybe Dark2 is not that good after all.

PR #244, opened 2020-07-05
Add slepc installtion to travis, fix links

michalk8:
TODO:
- test which way is best to speed up the tests (precomputed adatas vs slicing + trans. matrix normalization)
-  cache dependencies in travis


michalk8:
The tests run about in `45s` on my machine, using all cores (pytest-parallel is faster than xdist in our case).
 I've opted to cache the adata for speed (we can regenerate them if needed).
Once the SLEPc works, will merge.

Marius1311:
Very good idea to speed up our tests, will positively impact our review process.

Issue #248, opened 2020-07-06
Lineage class does not allow for np.mean


Marius1311:
I'm slicing a lineage class object to get values restricted to one cluster and I would like to compute the mean along axis 0, i.e. the mean probability per cluster:
`np.mean(adata.obsm['to_final_states'][mask, :], axis=0)`

```pytb
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
~/miniconda3/envs/cellrank/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

~/Projects/cellrank/cellrank/tools/_lineage.py in _repr_html_(self)
    419             )
    420         else:
--> 421             body = "".join(format_row(i) for i in range(self.shape[0]))
    422 
    423         cells = "cells" if self.shape[0] > 1 else "cell"

~/Projects/cellrank/cellrank/tools/_lineage.py in <genexpr>(.0)
    419             )
    420         else:
--> 421             body = "".join(format_row(i) for i in range(self.shape[0]))
    422 
    423         cells = "cells" if self.shape[0] > 1 else "cell"

~/Projects/cellrank/cellrank/tools/_lineage.py in format_row(r)
    389                 f"{super(type(self), self).__getitem__((r, c)):.06f}"
    390                 f"</td>"
--> 391                 for c in range(self.shape[1])
    392             )
    393             return f"<tr>{cells}</tr>"

IndexError: tuple index out of range
Lineage([0.09438334, 0.39487063, 0.51074603],
  names([goblet cells, ciliated cells, AT2 cells]))
```
So I guess it returns the right thing, but still complains?

#### Versions:
cellrank==1.0.0-b.3 scanpy==1.5.1 anndata==0.7.3 numpy==1.18.5 scipy==1.4.1 pandas==1.0.5 scikit-learn==0.22 statsmodels==0.10.2 python-igraph==0.7.1 scvelo==0.2.2.dev51+ga7de78a
> ...

michalk8:
I will have to modify __array_ufunc__ to make this work as intented. This will make it also nicer w.r.t. to the names and colors preserved.

michalk8:
Closed via #250

PR #250, opened 2020-07-07
Fix lineage repr

### LINKED ISSUE #249: Add pie chart plotting for the lineage class
Would be super nice if a lineage object would have a convenient way to visualise distributions, ie in pie charts. I'm thinking of a method `lineage.plot_pie_chart(mask: Optional[Iterable[bool]] = None)`, where mask is optional and could select a cluster. The method should compute the mean over all cells selected by the mask along axis 0 and plot something like the below:
image

where names and colors can be taken from the lineage object directly!

michalk8:
There's still a significan bug - the probabilites are not normalized to 1.

closes #249 #248


michalk8:
@Marius1311 I've changed the functionality of `Lineage` - I've modified function which operate on `Lineage` so that most of them should be shape-preserving, so beware.
Most importantly, the `html_repr` should be more robust - it works on transposed data as well (overall better handling there), + the reduction along 0th, 1st axis also display the function name.
I've also included the plotting fn, slightly modified (don't think we need the parameter you've suggested, we can always subset the object).
There's an issue with `.X` attribute - `similarity_plots` test fail if we don't copy the data (which we don't want to) - it's due to the fact that the rows don't sum to 1. I can't find the problem, if you have the time, I'd appreciate if you could look into it.

Marius1311:
Hi MIke, sorry, I don't really understand what the problem is here, I think it's best if you explain this to me tomorrow, then I will have a look at it.

michalk8:
@Marius1311 it's simple, somewhere, we have an `inplace` operation which modifies the values of lineage probabilities.
If using `.X` return a copy, everything's fine
I have a suspect: I've changed the reductions to not use `.copy` (seemed unnecesary - I've checked the code [but probably not thoroughly enough] and found no reason to do so): https://github.com/theislab/cellrank/pull/250/files#diff-afc45402aad45051cca8a4f21845bcf6L543
Since you've written it, maybe you know better.

michalk8:
I think I will create a separate issue for this afterwards.

michalk8:
Ok, going to merge this, tests pass. Will create a new issue regarding `.X`.

PR #255, opened 2020-07-08
Add tests for connectivity-based transition matrix

### LINKED ISSUE #219: Introduce symmetric normalization for transition matrices
At the moment, we row-normalize by just multiplying from the right. It's fine for directed matrices but it's a bit inconcistent for symmetric matrices (connectivities). See symmetric normalization in scanpy: https://github.com/theislab/scanpy/blob/a6590851dfbace6dc07b35b4f5d59d6bf9952446/scanpy/neighbors/__init__.py#L822

vs. our normalization:
https://github.com/theislab/cellrank/blob/164178875b60cd91dbc06a4d0eecb0407af1f35b/cellrank/tools/_utils.py#L741

Marius1311:

## Types of changes

## Description
This adds two tests that check whether our transition matrix, computed just based on connectivities, matches with scanpy's non-symmetric transition matrix. There is one test for doing this with and one without density normalization. Both pass on my machine. 

## How has this been tested?
This adds tests. 

## Closes
closes #219

### DIFF
--- a/tests/test_kernels.py
+++ b/tests/test_kernels.py
@@ -3,6 +3,7 @@ import numpy as np
 import pytest
 
 import anndata
+from scanpy import Neighbors
 
 from _helpers import (
     bias_knn,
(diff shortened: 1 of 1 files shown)


Marius1311:
@michalk8 , do we already have a utility that checks for equality of sparse matrices? I check separetely the shape, length of data and data itself by calling
```
        assert T_sc.shape == T_cr.shape
        assert len(T_sc.indices) == len(T_cr.indices)
        assert np.allclose((T_cr - T_sc).data, 0)
```

It's fine this way, just in case you already have a utility, we could use that.

PR #257, opened 2020-07-09
Add parallel to iterative_solver

### LINKED ISSUE #256: Paralellize computation of absorption probabilities
- [ ] Other?

For large problems, we use an iterative solver to get the absorption probabilities. This works by solving a series of linear problems: 
```python
            def flex_solve(M, B, solver=gmres, init_indices=None):
                # solve a series of linear problems using an iterative solver
                x_list, info_list = [], []
                x = None
                for ix, b in enumerate(B.T):
                    # use the previous problem solution as initialisation
                    if init_indices is not None:
                        x0 = None if ix in init_indices else x
                    else:
                        x0 = None
                    x, info = solver(M, b.toarray().flatten(), tol=tol, x0=x0)
                    x_list.append(x[:, None])
                    info_list.append(info)
```
As you can see, I loop over columns in the matrix `B` and solve one linear problem per column. This could easily be parallelized because these problems are independent. At the moment, I use the solution of one problem to initialize the next, but the speedup I get form that is marginal, so we can remove that to have fully independent problems, which should parallelize very well. It would be great if you could implement this and benchmark on one large example (lung) to see how big our speedup is here.

michalk8:

## Title
Parallelize solving of the system when computing absorption probabilities.

## Types of changes

## Description

Added an option to solve the system in parallel, default is still sequential.

## How has this been tested?
No unit tests yet - I will add one test using parallel option.

##Benchmarking
For now, only on pancres, 20 lineages.
Before:
1.12 s ± 23.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
After:
954 ms ± 14.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

## Closes
closes #256 

@Marius1311 
There was previously this line: `x0 = None if ix in init_indices else x`, but it doesn't seem correct: https://github.com/theislab/cellrank/blob/master/cellrank/tools/estimators/_base_estimator.py#L419. Am I wrong?
```

### DIFF
--- a/cellrank/tools/estimators/_base_estimator.py
+++ b/cellrank/tools/estimators/_base_estimator.py
@@ -324,6 +324,13 @@ class BaseEstimator(ABC):
         use_initialization
             Only relevant when using an iterative solver. In that case, the solution of absorbing states from the same
             recurrent class can be used as initialization to the iterative solver.
+        n_jobs
+            Number of parallel jobs. If `-1`, use all available cores. If `None` or `1`, the execution is sequential.
+        backend
+            Which backend to use for multiprocessing. Only used when :paramref:`n_jobs` `!=1`.
+            See :class:`joblib.Parallel` for valid options.
+        show_progress_bar
+            Whether to show a progress bar tracking. Only used when :paramref:`n_jobs` `!=1`.
 
         Returns
         -------
--- a/cellrank/tools/estimators/_base_estimator.py
+++ b/cellrank/tools/estimators/_base_estimator.py
@@ -16,23 +16,22 @@ from typing import (
 )
 from pathlib import Path
 
-import numpy as np
-import scipy
-import pandas as pd
-from pandas import Series
-from scipy.stats import entropy, ranksums
-from scipy.linalg import solve
-from scipy.sparse import issparse, spmatrix, csr_matrix
-from pandas.api.types import infer_dtype, is_categorical_dtype
-from scipy.sparse.linalg import eigs, gmres
-
 import matplotlib as mpl
 import matplotlib.cm as cm
 import matplotlib.pyplot as plt
 
 import scvelo as scv
 
+import numpy as np
+import scipy
+import pandas as pd
+from pandas import Series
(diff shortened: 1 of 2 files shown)


michalk8:
Also, you realize setting `._lin_probs` in this method overrides the computation done in `.{compute,set}_main_states()` in GPCCA?
Just checking whether it's intended.

Marius1311:
Hi Mike, what's the issue with `x0 = None if ix in init_indices else x`? Basically, I want to say that for cells that belong to the same metastable state, use the previous solution as initialization because these problems are likely to have similar solutions. If however the last cell came from a different metastable state, don't use this initialization. 

You are right there may be an issue here: cells aren't ordered by metastable states in the matrix `B`, ie I'm not sure whether I'm really making use of this initialization at all!

Marius1311:
The pancreas is too small though to benchmark this, can you try in lung or in the morris reprogramming data please? Because so far, there is no significant speed up.

Marius1311:
I think the first thing to check here would really be: in the old formulation without parallelization, did we make use of clever initialization? Because if not, we should change that, and see how much improvement that gives. Once we have that, we can benchmark this against a parallelized version.

Marius1311:
Okay, I think the `compute_absorption_probabilities` function needs a bit of work. Currently, it's too complicated. I will look into this hopefully later today, and refactor it. I will move the heavy lifting into a utility function we can test easily. Until then, I think it makes sense for you to pause your paralelization efforts.

michalk8:
> Hi Mike, what's the issue with `x0 = None if ix in init_indices else x`? Basically, I want to say that for cells that belong to the same metastable state, use the previous solution as initialization because these problems are likely to have similar solutions.

Ah, I see, thanks for the explanation.

> Until then, I think it makes sense for you to pause your paralelization efforts.

Agreed.

Marius1311:
I'm still on the refactoring, hopefully will be done tomorrow morning!

michalk8:
Thanks @Marius1311 , I'm happy to merge this!

PR #259, opened 2020-07-09
Refactor computation of absorption probabilities

Marius1311:

## Types of changes

## Description

This refactors the computation of absorption probabilities so that it's carried out in chunks by utility functions that we can easily test. We had a logical bug in there which meant that the initialization based on past problems was basically inefficient, so this could also lead to a speedup. 

## How has this been tested?

Need to add some tests.

### DIFF
--- a/cellrank/tools/_utils.py
+++ b/cellrank/tools/_utils.py
@@ -1403,3 +1403,60 @@ def _colors_in_order(
         color_list.append(adata.uns[f"{cluster_key}_colors"][mask][0])
 
     return color_list
+
+
+def _get_cat_and_null_indices(cat_series: Series):
+    """
+    Given a categorical pandas Series, get the indices corresponding to categories and Nans.
+
+    Parameters
+    --------
+    cat_series
+        Series that contains categorical annotations
+
+    Returns
+    cat_indices
+        numpy array containing the indices of elements corresponding to categories in `cat_series`
+    null_indices
+        numpy array containing the indices of elements corresponding to Nans in `cat_series`
+    lookup_dict
+        dict containing categories of `cat_series` as keys and a numpy array of corresponding indices as values
+    """
+
+    # check the dtype
+    if not (cat_series.dtype == "category"):
+        raise TypeError(
+            f"Expected `cat_series` to be categorical, found {cat_series.dtype}"
+        )
+
+    # define a dict that has category names as keys and arrays of indices as values
+    lookup_dict = {
+        cat: np.where(cat_series == cat)[0] for cat in cat_series.cat.categories
+    }
+    all_indices = np.arange(len(cat_series))
+
+    # collect all category indices
+    cat_indices = np.concatenate(list(lookup_dict.values()))
+
+    # collect all null indices (the ones where we have NaN in `cat_series`)
+    null_indices = np.array([*(set(all_indices) - set(cat_indices))])
+
+    # check that null indices and cat indices are unique
+    assert (
+        np.unique(cat_indices, return_counts=True)[1] == 1
+    ).all(), "Cat indices are not unique"
+    assert (
+        np.unique(null_indices, return_counts=True)[1] == 1
+    ).all(), "Null indices are not unique"
+
+    # check that there is no overlap
+    assert (
+        len(set(cat_indices).intersection(set(null_indices))) == 0
+    ), "Cat and null indices overlap"
+
+    # check that their untion is the set of all indices
+    assert set(cat_indices).union(set(null_indices)) == set(
+        all_indices
+    ), "Some indices got lost on the way"
+
+    return cat_indices, null_indices, lookup_dict
--- a/cellrank/tools/_utils.py
+++ b/cellrank/tools/_utils.py
@@ -17,9 +17,12 @@ from typing import (
 )
 from itertools import tee, product, combinations
 
+import matplotlib.colors as mcolors
+
 import numpy as np
 import pandas as pd
 from pandas import Series
+from cellrank import logging as logg
 from numpy.linalg import norm as d_norm
 from scipy.linalg import solve
 from scipy.sparse import eye, issparse, spmatrix, csr_matrix
@@ -28,10 +31,6 @@ from pandas.api.types import infer_dtype, is_categorical_dtype
 from sklearn.neighbors import NearestNeighbors
 from scipy.sparse.linalg import norm as s_norm
 from scipy.sparse.linalg import gmres, lgmres, gcrotmk, bicgstab
-
-import matplotlib.colors as mcolors
-
-from cellrank import logging as logg
 from cellrank.utils._utils import _get_neighs, _has_neighs, _get_neighs_params
 from cellrank.tools._colors import (
     _compute_mean_color,
@@ -1510,7 +1509,7 @@ def _solve_lin_system(
     tol
         Convergence threshold.
     use_eye
-        Solve `(I - may_a) * x = mat_b` instead.
+        Solve `(I - mat_a) * x = mat_b` instead.
 
     Returns
     --------
--- a/cellrank/tools/_utils.py
+++ b/cellrank/tools/_utils.py
@@ -17,9 +17,12 @@ from typing import (
 )
 from itertools import tee, product, combinations
 
+import matplotlib.colors as mcolors
+
 import numpy as np
 import pandas as pd
 from pandas import Series
+from cellrank import logging as logg
 from numpy.linalg import norm as d_norm
 from scipy.linalg import solve
 from scipy.sparse import issparse, spmatrix, csr_matrix
@@ -28,10 +31,6 @@ from pandas.api.types import infer_dtype, is_categorical_dtype
 from sklearn.neighbors import NearestNeighbors
 from scipy.sparse.linalg import norm as s_norm
 from scipy.sparse.linalg import gmres, lgmres, gcrotmk, bicgstab
-
-import matplotlib.colors as mcolors
-
-from cellrank import logging as logg
 from cellrank.utils._utils import _get_neighs, _has_neighs, _get_neighs_params
 from cellrank.tools._colors import (
     _compute_mean_color,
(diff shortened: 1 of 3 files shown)


Marius1311:
This is ready for your review. The main changes are:
- new utility functions
  - `_get_cat_and_null_indices`: this replaces the old way in which we exctracted the indices of recurrent and transient states
  - `_solve_lin_system`: solve sparse or dense linear systems using a direct or iterative solver
  - `_solve_many_sparse_problems`: solve linear system using iterative solver
- columns in the right hand side of the linear system are ordered now, such that all cells that belong to the same metastable state are next to another. This means the initialization should actually work now. I hope this did not break anything. 
- added a test function `compare_absorption_probabiliteis_with_reference` that compares with the outcome of `pydtmc` on a small toy example. This function may need your work: it's probably better to store the transition matrix somewhere else, so that it can be used by different test functions in the future. 

I haven't tested yet whether the initialization actually improves performance, will get to this.

This would benefit from further tests of the individual utilities if you find the time. At the moment, I only test the overall `.compute_absorption_probabilities` function.

Marius1311:
In case an iterative solver is being used, it would be really nice to have a progress bar, since we just loop over subproblems.

Marius1311:
Results on the lung (32k cells) using `gmres`: 
- without initialisation: 4:47 min
- with initialization: 4:51 min

So we may conclude that my "clever" initialization did not add anything in practice.

Marius1311:
## use_initialization = False
image

## use_initialization = True
image

We spend more time doing matrix vector products when we use the initialization, which means we need more iterations to converge, which means the initialization does not help.

Marius1311:
Tests are failing because there is no `pydtmc`. How would you like to handle this? We can either add it, or we can just hard-code the solution to the toy example.

cellrank/tools/_utils.py
```diff
@@ -1403,3 +1414,247 @@ def _colors_in_order(
...
+    return cat_indices, null_indices, lookup_dict
+
+
+def _solve_lin_system(
+    mat_a: Union[np.ndarray, spmatrix],
+    mat_b: Union[np.ndarray, spmatrix],
+    solver: str = "direct",
+    tol: float = 1e-5,
+    related_columns_in_b: Optional[Iterable] = None,
+    use_eye: bool = False,
```
michalk8, line 1485: @Marius1311 I've added this here because there was a code repetition in the caller in BaseEstimator. I hope you approve.

tests/test_cflare.py
```diff
@@ -260,6 +262,60 @@ def test_compute_lin_probs_keys_colors(self, adata_large: AnnData):
...
+                [0.0, 0.0, 0.0, 0.0, 0.05, 0.05, 0.0, 0.7, 0.2, 0.0, 0.0, 0.0],  # 6
+                [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0],  # 7
+                [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.8, 0.2, 0.0, 0.0, 0.0, 0.0],  # 8
+                [0.0, 0.0, 0.0, 0.0, 0.05, 0.05, 0.0, 0.0, 0.0, 0.0, 0.7, 0.2],  # 9
+                [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 0.0],  # 10
+                [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.8, 0.2, 0.0],  # 11
+            ]
+        )
+
+        absorption_probabilities_reference = np.array(
```
michalk8, line 285: Hardcoded the solution from PyDTMC.

cellrank/tools/estimators/_base_estimator.py
```diff
@@ -280,8 +280,7 @@ def compute_absorption_probabilities(
         self,
         keys: Optional[Sequence[str]] = None,
         check_irred: bool = False,
-        norm_by_frequ: bool = False,
-        use_iterative_solver: Optional[bool] = None,
+        solver: Optional[str] = None,
         tol: float = 1e-5,
         use_initialization: bool = True,
```
michalk8, line 285: @Marius1311 since you've found out it doesn't help, can we remove this?

michalk8:
I've just fixed the docstrings, types and removed some duplicated code + PyDTMC dependency. My only issue is this:
Btw, you no longer need to write: `DEBUG:` in logging (it does it for you)

Small request:
> @Marius1311 since you've found out it doesn't help, can we remove this?

Since it doesn't help, I'd remove this. Then I'm happy to merge and finish #257 (think it's important for benchmarking as well)

cellrank/tools/_utils.py
```diff
@@ -1478,41 +1482,45 @@ def _solve_lin_system(
...
     tol
-        Convergence threshold
+        Convergence threshold.
     related_columns_in_b
-        A list specifying columns in `mat_b` that we expect to have similar solutions. As an example, [0, 3, 9]
+        A list specifying columns in `mat_b` that we expect to have similar solutions. As an example, `[0, 3, 9]`
         would mean that columns 0, 1, 2 are expected to have similar solutions, then 3, 4, 5, 6, 7, 8, and then
         all remaining columns starting with 9.
+    use_eye
+        Solve `(I - may_a) * x = mat_b` instead.
```
Marius1311, line 1518: spelling mistake here.

Marius1311: should be `I - mat_a`

cellrank/tools/_utils.py
```diff
@@ -1526,18 +1534,19 @@ def _solve_lin_system(
...
+            logg.warning("Sparsifying `mat_a` for iterative solver.")
             mat_a = csr_matrix(mat_a)
         if not issparse(mat_b):
-            logg.warning("Sparsifying `mat_b` for iterative solver. ")
+            logg.warning("Sparsifying `mat_b` for iterative solver.")
             mat_b = csr_matrix(mat_b)
 
+        if use_eye:
+            mat_a = eye(mat_a.shape[0]) - mat_a
+
```
Marius1311, line 1549: I think this is fine.

cellrank/tools/_utils.py
```diff
@@ -1548,34 +1557,35 @@ def _solve_lin_system(
...
+            logg.warning("Densifying `mat_a` for direct solver.")
             mat_a = mat_a.toarray()
         if issparse(mat_b):
-            logg.warning("Densifying `mat_b` for direct solver. ")
+            logg.warning("Densifying `mat_b` for direct solver.")
             mat_b = mat_b.toarray()
 
+        if use_eye:
+            mat_a = np.eye(mat_a.shape[0]) - mat_a
+
```
Marius1311, line 1583: Yes, I think this is more elegant than what I had before.

cellrank/tools/_utils.py
```diff
@@ -1586,41 +1596,43 @@ def _solve_many_sparse_problems(
...
+    ------
     mat_a
         Matrix of shape `n x n`. We make no assumptions on `mat_a` being symmetric or positive definite.
     mat_b
         Matrix of shape `n x m`, with m << n.
     solver
-        Solver to use for the linear problem. Options are [gmres, lgmres, bicgstab, gcrotmk]. These
-        can be found in `scipy.sparse.linalg`.
+        Solver to use for the linear problem. Options are `['gmres', 'lgmres', 'bicgstab', 'gcrotmk']`.
+        These can be found in :func:`scipy.sparse.linalg`.
```
Marius1311, line 1619: Could also benchmark these different solvers, but I wouldn't spend much time on it, maybe just on a single problem to see whether there is a clear winner.

cellrank/tools/estimators/_base_estimator.py
```diff
@@ -370,15 +372,15 @@ def compute_absorption_probabilities(
...
+        # create Q (restriction transient-transient), S (restriction transient-recurrent)
         q = t[trans_indices, :][:, trans_indices]
         s = t[trans_indices, :][:, rec_indices]
 
         if check_irred:
             if self._is_irreducible is None:
                 self.compute_partition()
             if not self._is_irreducible:
-                logg.warning("The transition matrix is not irreducible. ")
+                logg.warning("The transition matrix is not irreducible")
```
Marius1311, line 383: I guess this is really picky, but further up, we terminated warning messages with a point. Shall we do the same here?

cellrank/tools/estimators/_base_estimator.py
```diff
@@ -388,19 +390,7 @@ def compute_absorption_probabilities(
...
-            logg.debug("DEBUG: Densifying matrices for direct solver. ")
-            q = q.toarray() if issparse(q) else q
-            s = s.toarray() if issparse(s) else s
-            eye = np.eye(len(trans_indices))
-        else:
-            if not issparse(q):
-                q = csr_matrix(q)
-            if not issparse(s):
-                s = csr_matrix(s)
-            eye = scipy.sparse.eye(len(trans_indices))
```
Marius1311: You're right, there was some repetition here.

tests/test_cflare.py
Marius1311: Fine by me.

cellrank/tools/estimators/_base_estimator.py
Marius1311: yes.

cellrank/tools/_utils.py
Marius1311: yes, I do.

michalk8:
I've included the changes in `parallelize` from https://github.com/theislab/cellrank/pull/257 and closed it (after accidentally merging it, but I've reverted it...) - you were right, creatin a new PR is the way to go.
I've also created a new issue, related to parallelization and would welcome your help there: https://github.com/theislab/cellrank/issues/260 (we might solve it trough PETSc).
After the tests pass, I am happy to merge this, thanks @Marius1311 !

PR #263, opened 2020-07-13
Improve heatmap

### LINKED ISSUE #262: Shaping the heatmap function
- [x] New plotting function in `cr.pl`?
- [ ] Other?

We have two functions to plot gene expression trends: `cr.pl.gene_trends` and `cr.pl.heatmap`. So far, I don't really use the latter because the former mostly does the job. However, I think there is one visualisation that's currently missing: something like scvelo's heatmap function:

image
See the dynamical modelling tutorial: https://scvelo.readthedocs.io/DynamicalModeling.html

This function simply takes a set of genes, convolves their expression over a window of a given length and sorts them in latent time. It would be very nice if we could adapt our heatmap function to do a similar thing while taking into account our lineage weights, i.e replacing the convolve with a model fit, as we already do it. 

I have tried producing such a plot with our current heatmap function:
image

There are a number of issues here:
- the genes aren't sorted using the time key, so I can't see the progression I see using scvelo.
- we use the same color map for all genes, which means that for most genes, I don't see anything because their variation is small compared to the others. 
- a bar at the top with cluster info would be nice
- a second bar at the top with the lineage weights would be nice

michalk8:

## Title
Add extra arguments to `cellrank.pl.heatmap`.

## Types of changes

## Description
Implements the changes as requested in #262. Added arguments are: `cluster_key`, `show_lineage_probabilities`, `n_convolve`.
Improved functionality of `cluster_genes` and `show_cbar`.

## How has this been tested?
Yes, I've added multiple unit tests for the arguments.

## Closes
closes #262


michalk8:
I have to merge this, there's a significal bug regarding the Lineage, see https://github.com/theislab/cellrank/pull/263/commits/e75eb8181eb2283c028aeaef74ee1b36ba14df25
@Marius1311 if you have any requests regarding the heatmap or there are any bugs, please create a new issue.

Marius1311:
okay, will do!

Issue #266, opened 2020-07-14
`g_fwd.set_main_states()` does not write absorption probs


mbuttner:
I compute absorption probs for a GPCCA estimator object and would like to write them to adata using `g_fwd.set_main_states()`. However, this writes something else and not the computed absorption probs. 

#### Versions:
cellrank==1.0.0-b.8 scanpy==1.5.1 anndata==0.7.3 numpy==1.17.4 scipy==1.4.1 pandas==0.25.2 scikit-learn==0.21.3 statsmodels==0.10.1 python-igraph==0.7.1 scvelo==0.2.1
> ...

michalk8:
For GPCCA estimator, `set_main_states` writes the GPCCA memberships to `g_fwd.lineage_probabilities`.
If you want the true absorption probabilities, you need to run `g_fwd.compute_absorption_probabilities` - this will override `g_fwd.lineage_probabilities`. @Marius1311 is still discussing which one we should use (most likely the latter).

michalk8:
closed via #276

Marius1311:
@mbuttner , we refactored our code quite a bit, will update tutorials soon. The workflow should be much clearer then!

PR #276, opened 2020-07-15
Unify workflow

### LINKED ISSUE #272: New anticipated workflow
- [ ] Other?

...
Both our estimators compute a set of metastable states, which in the case of `CFLARE` are all terminal and in the case of `GPCCA` may be initial, terminal or intermediate. Once metastable states have been computed using either estimator, we want to compute fate probabilities towards a subset of these, typically towards the terminal states. Also, we want to write annotations to AnnData, to have categorical annotations for initial and terminal states, and continious annotation for the probabilities of going towards the terminal states and coming from the initial states. 

Let's think of an elegant workflow for this, which is mostly shared by both estimators.

michalk8:

## Title
Unify stuff into more robust state, before #272 is addressed.

## Types of changes

## Description
Changed the following:
- GPCCA internal naming (_meta_states -> _premeta_states, _main_states -> _meta_states)
- new key for abs_prob: _abs_probs
- Move around plotting functions so that we can use `discrete` also for `.absorption_probabilities`

## How has this been tested?
It hasn't. The test will fail and need to be modified.

## Closes
closes #272 #277 #266 

## TODOs
Fix tests, ensure copying works, modify tutorials, version bump.

@Marius1311 tests will fail, but it will be greate if you could take a look over this and say whether you agree or not.


Marius1311:
I wouldn't rename `_meta_states` - they **are** metastable, just not terminal. What's the reasoning behind this?

Marius1311:
Okay, I understand that this is a quick fix and we will have to do this properly at a later time point. Feel free to merge if you think this solves conceptual issues for now.

michalk8:
I've been working on a cleaner and extensible implemetation, should be ready today.

Marius1311:
Awesome, looking forward!

Marius1311:
what's the state of this?

michalk8:
It took me while till I managed to get almost everything working (NYI reading from adata, copying) again under the refactoring. The tests will fail, but you can still give me feedback on the code if you want.
TODOs
- [ ] fix cloning
- [ ] fix reading from adata
- [ ] implement saving (pickle)
- [ ] update docstrings with docrep

Marius1311:
Hi Mike, does this already reflect the changes we discussed?

michalk8:
If you mean not normalizing the metastable states after subsetting + combining them, then yes.

Marius1311:
Okay, let me know when you think this is ready for my review.

michalk8:
@Marius1311 you can now review it. All the tests locally pass. I just need to write some for more for pickling + copying.

michalk8:
1 more thing: I opted for  not saving metastable states from GPCCA. Do you think it's a good idea? In this case, we can't reconstruct the object from adata, but unpickling/pickling still works.

Marius1311:
what do you mean by you're not saving them? You're not writing them to the estimator or the anndata object?

michalk8:
I don't write them to anndata object.

michalk8:
Looks like matplotlib 3.3.0 broke the plotting tests. I will address this tomorrow.

michalk8:
Nevermind, we had 1 bug in our tests + 1 in the heatmap, since I forgot to add `"absorption probabilities"` as ylabel and previously matplotlib didn't complain.

michalk8:
@Marius1311 why do we need `cellrank.tl.fin_{root,final}` when we have lineages? I refactored them to use `.fit` method, but it's inefficient now, since calling `cr.tl.lineages` after `cr.tl.find_final`, some computation is being repeated. If we want 2 separate methods, one that estimates only final states and the other also computes abs. probs. then I'd remove the raise from `cr.tl.lineages`. Other option is to split `.fit` into two parts or completely forgo `cr.tl.lineages` and add option `compute_abs_probs` or something. Right now I'm inclined towards the last option.

michalk8:
Also now addresses #277

Marius1311:
> @Marius1311 why do we need `cellrank.tl.fin_{root,final}` when we have lineages? I refactored them to use `.fit` method, but it's inefficient now, since calling `cr.tl.lineages` after `cr.tl.find_final`, some computation is being repeated. If we want 2 separate methods, one that estimates only final states and the other also computes abs. probs. then I'd remove the raise from `cr.tl.lineages`. Other option is to split `.fit` into two parts or completely forgo `cr.tl.lineages` and add option `compute_abs_probs` or something. Right now I'm inclined towards the last option.
> 
> Could you also please add some summary for GPCCA class, like CFLARE has - it's rather empty.

High level functions are indeed an issue - I will think about this later today.

michalk8:
@Marius1311 have you had a chance to go over it? I've fixed the `.fit` method as well as gone over the docs, they seem correct.

Marius1311:
Thanks Mike! I haven't gone through this yet, will do so today!

Marius1311:
Comments
- I got `ModuleNotFoundError` for `docrep`, which I then pip-installed
- I got ModuleNotFoundError` for `wrapt`, which I then pip-installed
- After installing both of these, I was able to import cellrank in a jupyter notebook
- upon running `g_fwd.compute_schur()`, I get `When computing metastable states, choose a number of states NOT in `[6, 10]``, which is great!

Marius1311:
Okay, more comments:
- `cellrank.tools._utils._create_root_final_annotations` needs some changes. This was a utility function I used to combine root and final categorical annotations. So far, this was acting on the level of estimator objects, however, it would be much better if this were to act on the level of an anndata object, once root and final annotations have been written. I already adjusted this and pushed a commit.

Marius1311:
Concluding, I think this is beautiful work and the only thing yet missing is proper handling of the `n_states=1` case.

michalk8:
@Marius1311 I've dealt with the 1 case just by saving the stationary distribution, see here: https://github.com/theislab/cellrank/pull/276/commits/72eaf405647ee50ec6d146e83caee193f4af3fcd
If you're fine with this, I think we can merge this after the tests pass. Afterwards, we need to update the tutorials.

Marius1311:
Thanks @michalk8, I will have a look at this, hopefully later today.

Marius1311:
Hi Mike, one conceptual thing: `root_states_probs` and `final_states_probs` should always be scaled to [0, 1] such that for every final/root state, there is at least one cell in `final/root_states_probs` which has a value of 1.0. For `n_s > 1`, this works already, i.e. in the pancreas for the final states, I can call

```
mask = adata.obs['clusters'] == 'Alpha'
print(adata[mask].obs['final_states_probs'].max())

mask = adata.obs['clusters'] == 'Beta'
print(adata[mask].obs['final_states_probs'].max())

mask = adata.obs['clusters'] == 'Epsilon'
print(adata[mask].obs['final_states_probs'].max())
```

and it prints `1.0` three times. However, this is currently not the case for `n_s=1` - the scaling is just missing. Please add this, such that also for a single root/final states, we scale such that the max cell is at 1.0. Would be nice to also have a test for this.

Marius1311:
Oh, I forgot to say why this is important: it's important when we interface to latent time because otherwise, scvelo does not recognize our root/final states.

Marius1311:
Other than that, everything seems fine.

michalk8:
The tests pass (it's not showing, because it's too slow, but I've checked all of them) - only 1 fails, due to some fatal error in msmtools (stationary distribution). Merging this.

Marius1311:
Okay, is it worthwhile to explore the msmtools error?

Marius1311:
Thanks for merging this, by the way!

PR #281, opened 2020-07-23
Refactor the VelocityKernel

### LINKED ISSUE #278: Velocity graph function in cellrank
- [ ] Other?

We should start thinking about how we want the cellrank velocity graph function to behave. On a high level, I think
- it should be a method of the `VelocityKernel`
- it should be called when we call `.compute_transition_matrix`
- it needs a pre-computed KNN graph to run. If None is present, it computes one using default values and issues a warning. No keywords can be passed to the KNN graph computation, this should really be done in `sc.pp.neighbors`
- it should take us all the way from velocity vectors and a KNN graph to transition probabilities, i.e. a combination of our old `.compute_transition_probabilities` and `scv.tl.velocity_graph` 
- it should support different modes, i.e. deterministic, stochastic based on the mean vector, stochastic based on 2nd order approximation, stochastic based on sampling.

Marius1311:

## Title
Provides a function to compute `velocity_correlations`. This moves the position in the pipeline at which we interface from scvelo to cellrank: whereas this was after `scv.tl.velocity_graph` before, it's one step earlier now, directly after `scv.tl.velocity`

## Types of changes

## Description
- refactor `.compute_transition_probabilities` from the `Kernel` class to not contain a bunch of stuff that is really just specific to the `VelocityKernel` and does not generalize to other kernels
- refactor the `.read_from_anndata` function from the `VelocityKernel` to read velocities and gene counts, rather than cosine correlations
- introduce utility functions to compute correlations between velocity vectors and transcirptomic displacements, `_pearson_corr` and `_cosine_corr`
- introduce a differentiable function `predict` which goes all the way from a velocity vector and a weight matrix to transition probabilities
- introduce uncertainty propagation 

## How has this been tested?
Tests yet missing

## Closes
closes #278


Marius1311:
The basic skelleton is there, I just have a scalabitliy issue now because I don't compute the hessian effectively yet. See below for just running this on 10 cells: 
image
Basically, all the time is taken by computing these Hessians.

Marius1311:
@michalk8 , I need your help here. We have a performance bottleneck here that stems from the computation of derivatives. In short, we need to compute the diagonal elements of a Hessian tensor of shape `(K, G, G)`, where `K` is the number of neighbors and `G` is the number of genes. We know from theory that for a general function, computing the diagonal elements of the Hessian is just as expensive as computing the full Hessian, so we're in an unfortunate spot where, even if I were to compute the derivatives for the diagonal elements analytically or using symbolic differentiation, this would not reduct the cost in comparison to computing the full matrix. So we may just as well compute the full hessian tensor.

I compute the full hessian tensor using Jax, which is as simple as calling `jacfwd(jacrev(_predict_fwd))`, where `_predict_fwd` is the function that gives us transition probabilities. Let's think of this `_predict_fwd` as a function `R^G -> R^K`. JAX traces the computational graph using abstract shaped arrays, i.e. `jacfwd(jacrev(_predict_fwd))`returns a function that gives you the hessian tensor for whatever inputs. Only if the shape of the underlying prediction function `f` changes, i.e. if `G` or `K` change, then the actual hessian function needs to be set up all again. Here is the **problem**: `K` changes in our case because we symetrise the KNN graph, so cells will not have exactly the same number of neighbors. 

In a typical example, where you have say 30 nearest neighbors, the actual `K` per call could lie in the range `(30, 120)`. That's a problems for JAX becasue everytime a new value of `K` is encountered, the computational graph needs to be traced again to get another function that returns hessian tensors... And this slows us down in practice.

To get around this, I tried to be clever and I decomposed `f` into a set of smaller functions `g, h, ...` which individually never change shape. I then used Jax to compute derivatives for these smaller functions and put the pieces back together to get the overall Hessian tensor. This works, i.e. I get the right hessian tensor **but** there is a new bottleneck. Let's call this approach here the semi-automatic approach, and Jax directly on `_predict_fwd` the automatic approach. The semi-automatic approach will be ~4x faster than the automatic approach if Jax needs to re-trace the computational graph, i.e. if `K` changes, but it will also be ~3x slower than the automatic approach if `K` does not change, i.e. if Jax just evaluates it's pre-compiled hessian function. 

I tried to improve this a bit by jitting parts of my semi-automatic approach, but the numbers don't change dramatically. My semi-automatic approach is still in a jupyter notebook because I think overall, when applied to a real dataset, it would be slower than the automatic approach. This is based on the following observation:
- the automatic approach takes 18 min on 2.5k cells for all computations (to get the final transition matrix)
- the semi-autmoatic approach takes 7s on 10 cells, so I reckon it would take 17.5 min on 2.5k cells. However, this only includes the time to compute the Hessians, nothing else)

So in this small case (pancreas), both approaches are still comparable, but when we go to higher cell numbers, the automatic approach will have an edge because once it has seen the problem for all possible values of `K`, it's 3x faster than the semi-autmoatic approach, and this tradeoff favours the automatic approach for increasing cell numbers. 

So what to do? Shall we just go for the autmatic approach? We could still talk about parallelization/comptuation of the GPU.

michalk8:
- Since it caches the graphs for different values of K, do you know the cache size (thinking maybe it's too small by default at some previous values are being evicted)?
- Would an approximation of Hessian be possible/feasible? I.e. here https://arxiv.org/pdf/2001.08355.pdf or here https://arxiv.org/pdf/2001.08355.pdf

If we're unable to find a solution on how to speed this up, I prefer the automatic solution, since a) will perform better on large data (still need to see this) b) it's more maintainable. Optionally, we could use both, one for smaller #cells, one for larger.  I can test the GPU/parallelization once I return.

Marius1311:
Thanks Mike! These are both very valid points. As to the first, I don't know what the size of the cache is, that would be important to look up. As to the second, you posted the same link twice. I had a look at it: this can be fast, but the error is only guaranteed to be as good as it is in a finite difference approximation, which can be pretty bad. Automatic differentiation is much more accurate here. The dangerous thing with finite differences is that these errors can accumulate and get quite large if you're not careful. Second, the authors don't provide an implementation. It would take us quite a while to implement any of these approaches, and even if we get it done, we would probably be slower than Jax for purely technical reasons, i.e. because Jax is very optimized. 

For now, I will finish the implementation using the automatic approach. Note that there is a mode parameter, so if the user does not want to wait for long, he is free to just use the deterministic version, which will be as fast as ever. 

GPU/parallelization would be great once you return. For now, enjoy your time at home!

Marius1311:
Good news @michalk8 : jitting the hessian function reduces compute time on the pancreas from 18 to 10 min. If I now change the order in which we iterate over cells, such that problems are ordered by dimensions, this could give a further substantial speedup.

Marius1311:
Okay, further speed ups. I changes the problem order, this brings us down to 10min for `mode=stochastic`. For `mode=deterministic`, I'm at 1s, which is pretty much what we had in sevelo.

Marius1311:
Tests are failing because Jax is not in the requirements.

Marius1311:
Replacing some `jnp` with standard `np` commands further decreased compute time to now 8:09 min with `mode=stochastic`

Marius1311:
Not jitting that hessian increases that time to 13:50 min.

Marius1311:
Changing the ordering of the problems, first the large, than small, further decreses the time to 7:46 min

Marius1311:
I decided to disable the density normalization for the velocity graph for now. To include it in the current framework, we need to change the setup a bit so that the density normalizing factors are taken into account directly in the `predict_transition_probabilities` function.

Marius1311:
Hi @michalk8 , this is ready for your review. A couple of points to note:
- tests are currently failing because Jax is not in our dependencies. Could you add it please?
- I disabled computing a density correction for the velocity kernel. This has technical reasons (has to be done a bit differently in the stochastic setup) as well as conceptual (not sure we actually want to correct for density here). I fixed our tests accordingly. 
- could you please add a progress bar to the `.compute_transition_probabilities` method? Would be really nice, esp. for the stochastic mode. However, also the other modes may take a little while on large datasets, so it would be nice to have for all of them. 
- I added 4 tests that check whether we match scvelo's output in the deterministic setup. All test are successful locally. 
- regarding speed, I pushed it to 7:40 min on the pancres using jitting and problem sorting. Further speedups could be achieved through parallelisation - however, I would delay this to a separate PR and focus on merging this quickly

michalk8:
@Marius1311 I've gone throught it, fixed some pet peeves, but the tests fail locally, with jax installed. Some of these are fixed in #276 (the new matplotlib version), but others are related to the changes you've made.
I'd also consider not requiring jax as hard dependency, since `jaxlib` is quite large + it slows down the imports.

michalk8:
I will fix the tests + regenerate the adata file to include velocity graph.

PR #315, opened 2020-08-19
Abs time moments

### LINKED ISSUE #308: absorption time mean and variance?
If a bit confused by the naming in the `.compute_absorption_probabilities` method of the base estimator. Why do we talk of time? We're not computing time until absorption, we're computing the probability that an individual cell will be absorbed by a given state, right?

michalk8:

## Title

Add more efficient computation of variance of time to absorption.
@Marius1311 I've settled on this, feel free to suggest otherwise:
- writing `'absorption_pseudotime'` and `'absorption_pseudotime_uncert'` to `adata.obs` + to `Estimator objcect` (previously were written only to estimator I think)
- I've kept the time 'til absorption + variance hidden under `_absorption_time_mean` and `_absorption_time_var` respectively

## Types of changes

## Description

+ some improvements to logging.

## How has this been tested?

I will add test later today which uses the normal equation.

## Closes

#closes #308

### DIFF
--- a/cellrank/tl/_linear_solver.py
+++ b/cellrank/tl/_linear_solver.py
@@ -423,15 +423,13 @@ def _solve_lin_system(
             use_petsc = False
 
     if use_eye:
-        if issparse(mat_a):
-            mat_a = speye(mat_a.shape[0]) - mat_a
-        else:
-            mat_a = np.eye(mat_a.shape[0]) - mat_a
+        mat_a = (
+            speye(mat_a.shape[0]) if issparse(mat_a) else np.eye(mat_a.shape[0])
+        ) - mat_a
 
     if solver == "direct" or n_jobs == 1:
-        logg.debug("Solving the linear system using direct matrix factorization")
-
         if use_petsc:
+            logg.debug("Solving the linear system directly using `PETSc`")
             return _petsc_mat_solve(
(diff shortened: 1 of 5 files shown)


Marius1311:
How do you transform the variance on the mean absorption time to give you the variance on the inverted mean absorption time?

michalk8:
I've been thinking whether it's correct to call it pseudotime - it's the time until absorption if we start at that given cell. I guess MFPT would be better suited for this - I think a formulat exists for it's variance so I will check it out later.

Marius1311:
Right, it's like a mean first passage time for a whole group of cells, i.e. all cells in the absorbing state.

Marius1311:
Only that it must be conditional, because a normal mean first passage time in an absorbing chain would be infinity. That's because it's the average over all possible paths that terminate in a given state. Some of these paths may be absorbed by other absorbing states, hence have infinite time. This would make the average infinite. So you have to condition to only average over those paths that actually terminate in your state of interest. So I would call this a conditional mean first passage time. 

Compare with https://www.tandfonline.com/doi/abs/10.1080/17509653.2013.783187?journalCode=tmse20

Marius1311:
Just read through the abstract, it's exactly that: a conditional mean first passage time.

PR #329, opened 2020-08-21
Automatic softmax scale

### LINKED ISSUE #328: Estimate softmax_scale for VelocityKernel
- [ ] Other?

As @Marius1311 had said, we can set it as we did previously by the median correlation. However, we'd have to run the method twice, it's better to estimate it using the fastest method (deterministic).

michalk8:

## Title

See below.

## Types of changes

## Description

Add estimation for `softmax_scale` when it's `None`. Also, fixed docs.

## How has this been tested?

I've added 1 test.

## Closes

closes #328

### DIFF
--- a/cellrank/ul/_docs.py
+++ b/cellrank/ul/_docs.py
@@ -24,7 +24,7 @@ backend
     Which backend to use for parallelization. See :class:`joblib.Parallel` for valid options."""
 _model = """\
 model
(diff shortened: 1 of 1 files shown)


Marius1311:
Fantastic, thanks. The default in scvelo for `scale` (which is the equivalent to `softmax_scale`) is 10, but I think that's too large for most datasets, so let's leave it at 4.

PR #356, opened 2020-09-01
Expose keyworded arguments for the trans_mat funciton

Marius1311:

## Title
Give more control in the high level transition matrix function

## Types of changes

## Description
This exposes some arguments from the underlying kernels in the high level transition matrix funciton

## How has this been tested?
Not at all.


Marius1311:
@michalk8 we could make this more general and expose kwargs from both the velo kernel and the connectivity kernel, what's your opinion?

michalk8 (approved):
I'd go for the following, as you already did, in VK: mode, backward_mode and softmax_scale
For CK: I think it's just density_norm, so no need to pack it under kwargs (if we introduce more arguments in the future, than sure).

michalk8:
I think this is fine. Merging.

Marius1311:
Great, thanks for looking into it

PR #437, opened 2020-10-15
Add p-values for gene drivers and 95% ci

### LINKED ISSUE #432: Return p-values for lineage driver genes
Hey guys, 

as I have already discussed with @Marius1311, it would be very useful to have p-values associated with the correlation scores returned for genes identified as lineage drivers. If I understand correctly, these are found by computing a rank correlation between their expression and the fate probabilities. I think a permutation test, or bootstrapping, could be some viable solutions for this.

michalk8:

## Types of changes

## Description

Add p-value calculation and CI for lineage drivers.

## How has this been tested?

No, needs to still be tested and old tests augmented.

## Closes

closes #432


Marius1311:
This is not needed at the moment and we will delay this until after our release.

Marius1311:
Hi Mike, please test this on the pancreas, for all 4 lineages, and report the values you get. I'm interested to see how many genes are significant when we consider FDR correction.

Marius1311:
Oh, and you do need to give more detail of how this CI was computed. Read through the wikipedia page, try to understand, quickly summarise in the docstring and ideally, cite! I'm sure this just didn't fall from heaven!

Marius1311:
Do we have tests which ensure that this does not change **anything** to previously computed p-values? This is top priority, needs to be tested on some example cases.

Marius1311:
Hi Mike, please update me on the progress here.

michalk8:
It's essentially done, except for 3 things:
- [ ] thoroughly check the formulas (should be fine, but better safe than sorry)
- [ ] improve the docs (add reference to the formulas used)
- [ ] write tests

Also, after using the FDR correction (BH), none of the genes for any lineage (Alpha, Beta, Epsilon) were found significant.

michalk8:
As for the progress on the tradeSeq statistical tests, I've had a look into their impl. - it's convoluted and would do a separate PR for that, since it will require much more effort than this - we will need to discuss this later (Thursday).

Marius1311:
> It's essentially done, except for 3 things:
> 
> * [ ]  thoroughly check the formulas (should be fine, but better safe than sorry)
> * [ ]  improve the docs (add reference to the formulas used)
> * [ ]  write tests
> 
> Also, after using the FDR correction (BH), none of the genes for any lineage (Alpha, Beta, Epsilon) were found significant.

Can you check please how scanpy does the multiple testing correction for differential expression testing? (`rank_marker_genes`)

Marius1311:
> As for the progress on the tradeSeq statistical tests, I've had a look into their impl. - it's convoluted and would do a separate PR for that, since it will require much more effort than this - we will need to discuss this later (Thursday).

Yes, let's do this in a separate PR.

Marius1311:
Next steps:
- use z-score transfomation before computing t-statistic
- try Fisher-transrormation
- permutation test

in this order. 
And: add Delta lineage for control.

Marius1311:
Hi Mike let's go though your progress on this tomorrow, excited to merge this!

Marius1311:
nice, thanks Mike. Can we merge later today?

michalk8:
We still need tests. I also removed `return_drivers`, but am thinking of reintroducing it, because I'm also saving the q-values in adata.var (or adata.raw.var). What do you think?
After the tests, we can merge - I've update requirements so that our tests will pass in the futures, but for official release on pypi, I'd remove this (since we really don't need h5py, it's scanpy's req).

michalk8:
I've also added some filtering options in `pl.lineage_drivers` - to filter out anything insignificant + the q-values are in the titles as well, if no title is specified.

michalk8:
@Marius1311 do you really want to merge this today or can it wait till tomorrow?

michalk8:
@Marius1311 could you please check the Fischer transform [here](https://github.com/theislab/cellrank/pull/437/files#diff-2e8fea22d3ec5969254f0ca1081db7b40dc4e382a3449cd6d92b9911fe3e2149R515)? I'd like to merge this today.
Afterwards, we can merge it - the tests pass except (Python3.7 received SEGFAULT, not sure why, will investigate later).

cellrank/tl/_utils.py
```diff
@@ -321,39 +332,224 @@ def _bias_knn(
...
+    corr = _mat_mat_corr_sparse(X, Y) if issparse(X) else _mat_mat_corr_dense(X, Y)
+
+    if method == TestMethod.FISCHER:
+        mean, se = np.arctanh(corr), 1.0 / np.sqrt(n - 3)
+        z_score = (np.arctanh(corr) - np.arctanh(0)) * np.sqrt(n - 3)
+
+        z = norm.ppf(0.975)
+        corr_ci_low = np.tanh(mean - z * se)
+        corr_ci_high = np.tanh(mean + z * se)
+        pvals = 2 * norm.cdf(-np.abs(z_score))
```
Marius1311, line 530: The maths seems correct, but would greatly benefit from some comments so we can still understand what we did in 4 weeks.

cellrank/tl/_utils.py
```diff
@@ -321,39 +332,224 @@ def _bias_knn(
...
+    X: Union[np.ndarray, spmatrix],
+    Y: np.ndarray,
+    method: TestMethod = TestMethod.FISCHER,
+    n_perms: Optional[int] = None,
+    seed: Optional[int] = None,
+    **kwargs,
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+    """
+    Compute the correlation between rows in matrix ``X`` columns of matrix ``Y``.
+
```
Marius1311, line 481: Either here or somewhere else do we need to give a bit more meat: how do we compute these p-values and confidence intervals, where do the formulas come from? Need to cite something.

cellrank/tl/_utils.py
Marius1311: e.g. why is z defined the way it is?

Marius1311: This clearly depends on the size of the confidence interval (which should, by the way, probably be a method parameter)

Marius1311 (approved):
This is fine, except for commenting your code and giving some more details in the doc string.

cellrank/tl/_utils.py
michalk8: Ok, I will expose the width of the CI (didn't want to overwhelm the user, but it might be useful).

cellrank/tl/_utils.py
michalk8: Ican cite the original paper on Fischer transformation (from 1920s) to be shown in the docs.
If you want this internal only, I'd just add the link to wikipedia and that's it.

Marius1311:
Amazing, thanks for merging this!

Issue #445, opened 2020-11-03
'Latest additions' section in the docs


Marius1311:
- [ ] Additional function parameters, change functionality or change defaults?
- [ ] New estimator in `cr.tl.estimators`?
- [ ] New kernel in `cr.tl.kernels`?
- [ ] New gene trend model in `cr.ul.models`?
- [ ] New plotting function in `cr.pl`?
- [x] Other?

I think it would make sense to have a scanpy-like "latest addition" section prominently placed in the `index.rst` so that people are informed about new features, like to upcoming p-values. Not many people will check the release notes, so all of the nice new features risk being overlooked otherwise.

michalk8:
Closed via #448

Marius1311:
Looks very nice, thanks.

PR #447, opened 2020-11-05
Implement custom VK functions

### LINKED ISSUE #330: Allow custom function for Velocity kernel
- [ ] Other?

Currently, it's correlation + softmax, but it might be useful to allow others to experiment with their impl. @Marius1311 what do you think?

michalk8:

## Title

Add new similarity schemes for VelocityKernel + accept custom user-defined function.

## Types of changes

## Description

Add similarity schemes for VelocityKernel according to: https://www.biorxiv.org/content/10.1101/2020.09.19.304584v1.full.pdf

## How has this been tested?

Not yet.

TODOs:
- [ ] populate docstrings (+ add more info for developers on whan inputs we provide/outputs we expect)
- [ ] rename the arguments (e.g. v for velocity vector/matrix, D for displacement matrix)
- [ ] remove stateful info from the classes (hope this fixes the SEGFAULT)
- [ ] find the test that fails and the other that SEGFAULTs the interpreter (see above)
- [ ] write tests (bunch of them)

@Marius1311 your input is needed here:
- [ ] handle "correlations" for cosine case*
- [ ] handle "correlations" for dot product case*
- [ ] handle "correlations" for custom fn *, **
- [ ] in the paper, they mention `g(x) is a bounded, positive, and non-decreasing function` - `g` is applied to the output of correlation/cosine sim./dot product (in our case `g` is identity) - any thoughts on this?

\* - could rename them to logits, since this term generalizes it
\*\* - we could also disallow them (though if we require only logits, then it's a very small impl. burden for devel)

## Closes

closes #330


Marius1311:
Hi Mike, this is something we should discuss on Wednesday. Thanks for kicking this off already! One question: wouldn't `g` be the softmax in our case? 

I will go through this part in the preprint more carefully so that we can discuss on Wednesday.

michalk8:
@Marius1311 could you please review the functions to asses they are implemented correctly?
Namely:
-https://github.com/theislab/cellrank/pull/447/files#diff-955d7bd9b19c6ac54b6d0d65fd4a5fcd496b34d470f28f002637f7bad061e936R41
- https://github.com/theislab/cellrank/pull/447/files#diff-955d7bd9b19c6ac54b6d0d65fd4a5fcd496b34d470f28f002637f7bad061e936R96
- https://github.com/theislab/cellrank/pull/447/files#diff-955d7bd9b19c6ac54b6d0d65fd4a5fcd496b34d470f28f002637f7bad061e936R225 (+2 function below)

Lastly, I think we should add them to docs somewhere in `Classes` section, since I will heavily make references to the paper - people will be able to know which scheme to choose and how to program their own.

michalk8:
I've disabled macOS and the tests are running, once they pass + you review the changes, I will merge it.
Could you please take care of https://github.com/theislab/cellrank/issues/445? I think you can formulate this much better than I do.
After #445 is merged, let's make 1.1.0 release.

cellrank/tl/kernels/_velocity_schemes.py
```diff
@@ -0,0 +1,345 @@
...
+
+
+try:
+    import jax.numpy as jnp
+    from jax import jit, hessian
+
+    @jit
+    def _softmax_jax(x: np.ndarray, softmax_scale) -> np.ndarray:
+        numerator = x * softmax_scale
+        numerator = jnp.exp(numerator - jnp.max(numerator))
```
Marius1311, line 28: Why is there the `-jnp.max(numerator)`?

cellrank/tl/kernels/__init__.py
```diff
@@ -2,5 +2,11 @@
 from cellrank.tl.kernels._base_kernel import Kernel, Constant
 from cellrank.tl.kernels._palantir_kernel import PalantirKernel
 from cellrank.tl.kernels._velocity_kernel import VelocityKernel
+from cellrank.tl.kernels._velocity_schemes import (
+    CosineScheme,
+    DotProductScheme,
+    CorrelationScheme,
+    SimilaritySchemeABC,
```
Marius1311, line 9: I take it this is the base scheme?

cellrank/tl/kernels/_velocity_kernel.py
```diff
@@ -17,23 +17,21 @@
...
-    _predict_transition_probabilities_jax_H,
-    _predict_transition_probabilities_numpy,
 )
 from cellrank.tl.kernels._base_kernel import (
     _RTOL,
     _LOG_USING_CACHE,
     _ERROR_EMPTY_CACHE_MSG,
     AnnData,
 )
+from cellrank.tl.kernels._velocity_schemes import Scheme, _get_scheme
```
Marius1311, line 34: `Scheme` is the base class?

cellrank/tl/kernels/_velocity_kernel.py
```diff
@@ -99,7 +97,7 @@ def __init__(
         self._vkey = vkey  # for copy
         self._xkey = xkey
         self._gene_subset = gene_subset
-        self._pearson_correlations = None
+        self._logits = None
```
Marius1311, line 100: Makes a lot of sense to rename to `_logits` in my mind.

cellrank/tl/kernels/_velocity_kernel.py
```diff
@@ -502,15 +521,19 @@ def _run_stochastic(
...
+                )
 
             # compute zero order term
-            p_0, c = _predict_transition_probabilities_numpy(
-                v[None, :], W, softmax_scale=softmax_scale
-            )
+            p_0, c = scheme(v[None, :], W, softmax_scale)
 
             # compute second order term (note that the first order term cancels)
             p_2 = 0.5 * H_diag.dot(variance[ix])
```
Marius1311, line 539: One could mention somewhere that our formulas for the expected transition probabilities hold as long as we assume gene-wise velocities to be independent, which is the case in the current scvelo framework. If this ever gets extended, we need to re-work our uncertainty propabation.

cellrank/tl/kernels/_velocity_schemes.py
Marius1311: This is the softmax: https://en.wikipedia.org/wiki/Softmax_function

I really don't get this subtracted term, but realised that we had included it before. Why?

cellrank/tl/kernels/_velocity_schemes.py
```diff
@@ -0,0 +1,345 @@
...
+    ):
+        if center_mean:
+            # pearson correlation, otherwise cosine
+            W -= W.mean(axis=1)[:, None]
+            X -= X.mean()
+
+        if scale_by_norm:
+            denom = jnp.linalg.norm(X) * jnp.linalg.norm(W, axis=1)
+            mask = jnp.isclose(denom, 0)
+            denom = jnp.where(jnp.isclose(denom, 0), 1, denom)  # essential
```
Marius1311, line 57: So the denominator is zero if either we have zero velocity or zero displacement. In that case, we set the denominator to 1. What's the intuition behind that? If I have a zero velocity vector for a given cell, wouldn't that just mean it stays where it is? Not sure.

cellrank/tl/kernels/_velocity_schemes.py
```diff
@@ -0,0 +1,345 @@
...
+    scale_by_norm: bool = True,
+) -> Tuple[np.ndarray, np.ndarray]:
+    if center_mean:
+        # pearson correlation
+        W -= np.expand_dims(np_mean(W, axis=1), axis=1)
+
+    if X.shape[0] == 1:
+        if center_mean:
+            # pearson correlation
+            X = X - np.mean(X)
```
Marius1311, line 111: Why is this different here than in the Jax case?

Marius1311: Is that because this works for the backwards case as well?

Marius1311 (commented):
Very nice work @michalk8. Just a couple of open questions that I would appreciate your input on.

Marius1311:
Comments:
 - expose `scheme` in high level transition matrix function.

Marius1311:
I looked into this on the pancreas here: https://github.com/theislab/cellrank_notebooks_dev/pull/90

cellrank/tl/kernels/_velocity_schemes.py
michalk8: It's for numerical stability, nothing else.

cellrank/tl/kernels/__init__.py
michalk8: Yes, it's the abstract base class (ABC) from which users should derive their subclasses.
I didn't want to expose `SimilarityScheme`, because that's what we use (it also includes mandatory hessian function).

cellrank/tl/kernels/_velocity_kernel.py
michalk8: `Scheme` is the basis for `Cosine`, `Correlation` and `DotProduct`, which also includes the hessian (it has somewhat speciic constructor).

cellrank/tl/kernels/_velocity_schemes.py
michalk8: If you refer to the `np_mean`, it's because in numba, `axis` cannot be specified in `np.mean`.
If you refer to the `X = X - ...` (instead of `X -= ...)`, it's because I can't use inplace assignment there.

cellrank/tl/kernels/_velocity_schemes.py
michalk8: Re stability:
```python3
import numpy as np
def _softmax(x: np.ndarray, softmax_scale: float) -> np.ndarray:
    numerator = x * softmax_scale
    numerator = np.exp(numerator)
    return numerator / np.sum(numerator)


def _softmax_stable(x: np.ndarray, softmax_scale: float) -> np.ndarray:
    numerator = x * softmax_scale
    numerator = np.exp(numerator - np.max(numerator))
    return numerator / np.sum(numerator)

np.random.seed(42)
_softmax(np.random.normal(size=(10,), loc=1000), 100)
# array([nan, nan, nan, nan, nan, nan, nan, nan, nan, nan])
np.random.seed(42)
_softmax_stable(np.random.normal(size=(10,), loc=1000), 100)
# array([1.68314704e-34, 1.33570495e-34, 7.25077912e-04, 1.81715293e-97,
       2.75338599e-89, 8.55578021e-39, 2.31953042e-58, 9.99274922e-01,
       8.26251375e-54, 1.03878654e-75])
```

michalk8:
Thanks for looking into it, I didn't expect much to change.
Finally merging this.

cellrank/tl/kernels/_velocity_schemes.py
Marius1311: got it. Please add a comment - when working together on software, it's just much easier when these non-entirely obvious things are commented on. 
Definitely a good idea to use a stable version of the softmax here!

cellrank/tl/kernels/__init__.py
Marius1311: okay.

cellrank/tl/kernels/_velocity_kernel.py
Marius1311: okay, thanks.

cellrank/tl/kernels/_velocity_schemes.py
Marius1311: @michalk8 ?

cellrank/tl/kernels/_velocity_schemes.py
Marius1311: Ah not I was thinking of why we have the extra if statement checking for `X.shape[0] == 1`, and I do assume that's because this function can handle the backward case as well?

cellrank/tl/kernels/_velocity_schemes.py
michalk8: @Marius1311 the masking is done to get the gradients flowing (they can't flow through NaN).
In case of 0 velocity vector, we will then get uniform probabilities across the neighbors (not sure if cells `x` is considering itself as a neighbour in the velo graph, but if yes, we could change it so that the self transition is 1).
In any case, the values in the numerator are ignored when computing the softmax: https://github.com/theislab/cellrank/pull/447/files/5b3a3bf113b77255de5e2676089f68414766de08#diff-955d7bd9b19c6ac54b6d0d65fd4a5fcd496b34d470f28f002637f7bad061e936R91
(i.e. in the case of displacement, if there are cells that are the same [displ. 0, but possible different velocities], the transition probability to those will be 0).

Marius1311: Hi Mike, this makes sense. Cell `x` is not considered to be a neighbor of itself. The current behaviour (uniform over neighbors if velocity vector is zero) makes sense to me. It won't matter in practice as this should be a 0-probability event and only happen in exceptional cases. The other case, zero displacement, also makes sense to me. Fine!

Issue #465, opened 2021-01-21
Using pyGPCCA under the hood


Marius1311:
- [ ] Additional function parameters, change functionality or change defaults?
- [ ] New estimator in `cr.tl.estimators`?
- [ ] New kernel in `cr.tl.kernels`?
- [ ] New gene trend model in `cr.ul.models`?
- [ ] New plotting function in `cr.pl`?
- [x] Other?

What will the process look like of switching to pyGPCCA under the hood? When can we initiate the move?

michalk8:
> What will the process look like of switching to pyGPCCA under the hood? When can we initiate the move?

I can start now, but merge only after pyGPCCA is on PyPI - we also need to discuss write release notes:
- mention circular embedding
- mention pyGPCCA

Marius1311:
Oh, now that you mention it, I just checked our release notes [here](https://cellrank.readthedocs.io/en/latest/release_notes.html) and we don't seem to have any for 1.1 - what happened there?

michalk8:
closed via #472

Marius1311:
Very nice!

Issue #453, opened 2020-11-23
Terminal states D-orthogonality error


le-ander:
We have briefly discussed this error before and from what I got, this should be about to be fixed anyways. I just wanted to leave this here for tracking and maybe to get a time estimate from you guys when you think this would be fixed?
Thanks a lot!

```python
cr.tl.terminal_states(adata, cluster_key='louvain_separate_fine', n_states=None, weight_connectivities=0.2)
```

```pytb
[[ 1.00000000e+00 -2.04305480e-03  1.71689486e-13]
 [-2.04305480e-03  1.00000000e+00  1.76311602e-16]
 [ 1.71687968e-13  1.76555548e-16  1.00000000e+00]]
WARNING: Using `3` components would split a block of complex conjugates. Increasing `n_components` to `4`
[[ 1.00000000e+00 -2.04305480e-03  1.01769287e-13  3.91856686e-14]
 [-2.04305480e-03  1.00000000e+00 -3.79125171e-16  3.71041088e-16]
 [ 1.01772757e-13 -3.79064185e-16  1.00000000e+00 -4.63171168e-16]
 [ 3.91856686e-14  3.70986878e-16 -4.61436445e-16  1.00000000e+00]]

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/opt/python/lib/python3.8/site-packages/cellrank/tl/estimators/_decomposition.py in compute_schur(self, n_components, initial_distribution, method, which, alpha)
    447         try:
--> 448             self._gpcca._do_schur_helper(n_components)
    449         except ValueError:

/opt/python/lib/python3.8/site-packages/cellrank/_vendor/msmtools/analysis/dense/gpcca.py in _do_schur_helper(self, m)
   1080         else:
-> 1081             self.X, self.R, self.eigenvalues = _do_schur(self.P.copy(), self.eta, m, self.z, self.method)
   1082 

/opt/python/lib/python3.8/site-packages/cellrank/_vendor/msmtools/analysis/dense/gpcca.py in _do_schur(P, eta, m, z, method, tol_krylov)
    257         print(X.T.dot(X*eta[:, None]))
--> 258         raise ValueError("Schur vectors appear to not be D-orthogonal.")
    259 

ValueError: Schur vectors appear to not be D-orthogonal.

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-18-e6c35a0187c4> in <module>
[...]

PR #472, opened 2021-02-01
pyGPCCA integration

### LINKED ISSUE #168: Tests for msmtools/GPCCA
These would be very nice - it would reduce my fear breaking everything... See https://github.com/msmdev/msmtools/issues/4

### LINKED ISSUE #469: Missing jax installation and stochastic mode
@Marius1311 I remember that an error popped up when trying `mode='stochastic'`, but jax has not been installed and that
this error was rather ominous.
This must've changed  in 1.1.0, when I've changed the detection (i.e. `'stochastic'` mode does not "need" jax, it just needs an attribute called `hessian` ([here](https://github.com/theislab/cellrank/blob/master/cellrank/tl/kernels/_velocity_kernel.py#L217)'s the check).
I did add dummy functions [here](https://github.com/theislab/cellrank/blob/master/cellrank/tl/kernels/_velocity_schemes.py#L73) which should produce a nice error message, but they don't.

#### Versions:
cellrank==1.1.0

michalk8:

## Title

Adds pyGPCCA as a dependency.

## Types of changes

## Description

Also updates pre-commit and fixes some docstrings, fixes #469  and fixes `nbviewer` CSS.

## Closes

closes #168 #453 #465 #469

TODOs:
- [x] verify that we're reading the correct attributes
- [x] search for and remove all previous `msmtools` things
- [ ] ~~in `compute_schur`, we say which clusters not to pick, maybe it would be better to say otherwise~~ I think this can be too much "overwhelming" info (i.e. the set of states we can't select is small than those we can), so I'd keep it as it is.
- [x] @Marius1311 discuss how to handle this change: (now) sparse matrix + brands = error; (before) ... = densify; currently, we issue a warning and densify (I think it's ok), alt. is to have a e.g. `cr.settings.densify = False`, which prevents doing this


Marius1311:
Hi @michalk8, re the last point, I would vote for warning and densifying. The reason is that this case will be **very** common: our matrices are basically always sparse, and many people, at least when initially installing, will not have SLEPSc, hence they are in exactly the situation you described.

cellrank/datasets/_datasets.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
Marius1311, line 1: Maybe just tell me once why you removed all of these?

cellrank/pl/_graph.py
```diff
@@ -174,7 +173,7 @@ def plot_arrows(curves, G, pos, ax, edge_weight_scale):
                 X, Y = X[::-1], Y[::-1]
 
             mid = len(X) // 2
-            posA, posB = zip(X[mid : mid + 2], Y[mid : mid + 2])  # noqa
```
Marius1311, line 177: What does the # noqa do?

cellrank/tl/kernels/_velocity_kernel.py
```diff
@@ -214,13 +213,19 @@ def compute_transition_matrix(
...
+                    scheme, "hessian"
+                ):  # this can also raise in our velocity scheme definition
+                    raise NotImplementedError
+            except NotImplementedError:
+                logg.warning(
+                    f"Unable to detect a method for Hessian computation. If using predefined functions, consider "
+                    f"installing `jax` as `pip install jax jaxlib`\n"
+                    f"Defaulting to `mode={VelocityMode.MONTE_CARLO.s!r}` and `n_samples={n_samples}`"
+                )
+                mode = VelocityMode.MONTE_CARLO
```
Marius1311, line 228: I guess this fixed the Jax error I was experiencing?

Marius1311 (approved):
Mike, this looks very well done. Did you have any troubles with tests or did they pass straight away?

cellrank/pl/_graph.py
michalk8: It's used to silence flake8

cellrank/datasets/_datasets.py
michalk8: They were only supposed to be used for Python2.

cellrank/tl/kernels/_velocity_kernel.py
michalk8: Yes.

michalk8:
Thanks for the review, merging.

cellrank/datasets/_datasets.py
Marius1311: ok, got it.

cellrank/pl/_graph.py
Marius1311: Okay, that's what I was guessing. I suppose the new tags silence only particular flake8 complaints.

cellrank/tl/kernels/_velocity_kernel.py
Marius1311: amazing, thanks.

Marius1311:
nice!

PR #502, opened 2021-02-19
Move priming deg

### LINKED ISSUE #500: Switch differentiation potential
- [ ] Other?

I would switch our differentiation potential (simple entropy over fate-probs) for STEMNETs `priming_degree` that we have already implemented in the circular projection (https://cellrank.readthedocs.io/en/latest/api/cellrank.pl.circular_projection.html#cellrank.pl.circular_projection). It has the conceptual advantage of monotonically increasing from progenitors to terminally differentiated cells.

michalk8:

## Types of changes

## Description

Add new properties: `cellrank.tl.Lineage.priming_{degree,direction}`, switch diff. potential (entropy) to priming degree.
Also fixes priming degree for terminal states (was NaN, is 1 now) and some logging errors.

## How has this been tested?

I've added 2 new tests for the above mentioned properties.

## Closes

closes #500


michalk8:
@Marius1311 - could you check if it works as expected? I've changed also the estimator property (`diff_potential` -> `priming_degree`).

Marius1311:
Fantastic, thanks Mike, will check today.

Marius1311:
This doesn't yet work as expected, I need to look into this a bit more tomorrow to understand what's going wrong here.

Marius1311:
Okay, I needed some time to think about this, here's what I came up with: 

I think the whole priming degree/diffusion potential thing is difficult enough that we should devote an own method to this at the level of the base estimator class as that's something you can do with lineage probabilities, regardless of how they have been computed. I have in in mind a method `.compute_lineage_priming` with parameters

- `method`: one of `entropy` or `kl_divergence`, where the former refers to palantir-like diffusion potential and the latter refers to STEMNET like priming degree
- `clusters`: a list of "early" clusters that are used to define the "average" lineage probabilities when using `kl_divergence`. By default, all cells are considered

This method should return a vector, normalised to [0, 1], where 0 corresponds to no priming and 1 corresponds to fully committed. This way, we could unify both mechanisms to compute lineage priming under a single method with unified interface and consistent output, while giving a bit more control to the user. I would then stop automatically computing this when computing absorption probabilities and I would also stop automatically plotting this when plotting absorption probabilities. This would avoid overloading these methods with extra parameters. 

In my experience, computing these measures can be a bit tricky and both the priming degree and the diffusion potential worked sometimes and sometimes not - that's why I believe we need to give a bit more control through the optional cluster selection. 

We could then have a dedicated plotting method `.plot_lineage_priming` - however, I think that's an overkill and it's fine to just write to `adata.obs`, from where this can be plotted conveniently using scanpy/scvelo. 

What's your opinion?

Marius1311 (changes requested):
See my comment, I think we need an extra method for this.

cellrank/tl/_lineage.py
```diff
@@ -706,6 +706,22 @@ def copy(self, _="C") -> "Lineage":
     def __copy__(self):
         return self.copy()
 
+    @property
+    def priming_degree(self) -> np.ndarray:
+        """Degree of priming as described in [Velten17]_."""
+        probs = self.X
+        return np.nan_to_num(
+            np.sum(probs * np.log2(probs / np.mean(probs, axis=0)), axis=1),
```
michalk8, line 714: @Marius1311 2 questions, both regarding:
Do you refer to this?
> clusters: a list of "early" clusters that are used to define the "average" lineage probabilities when using kl_divergence. By default, all cells are considered

michalk8 (commented):
@Marius1311 what about `normalize_by_mean` in `cellrank.pl.circular_projection`?
Should this also be part of the `calculate_priming_degree` method? (I think it should).

cellrank/tl/_lineage.py
michalk8: Also, I'd consider passing a vector of cell ids instead of cluster (more general, easier to implement - no need for cluster_key + checks).

michalk8:
@Marius1311 ready for review. Here are some TODOs:
- [ ] update docstrings (I again leave this up to you, think you can do it better)
- [ ] determine whether `early_cells` should also be exposed in `cellrank.pl.circular_projection`
- [ ] write tests for `BaseEstimator.compute_priming_degree` function

cellrank/tl/estimators/_base_estimator.py
```diff
@@ -317,8 +317,8 @@ def compute_absorption_probabilities(
         Compute absorption probabilities of a Markov chain.
 
         For each cell, this computes the probability of it reaching any of the approximate recurrent classes defined
-        by :paramref:`{fs}`. This also computes the entropy over absorption probabilities, which is a measure of cell
-        plasticity, see [Setty19]_.
+        by :paramref:`{fs}`. This also computes the priming degree as defined in [Velten17]_, see
```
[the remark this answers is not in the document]
Marius1311: sure!

Marius1311:
Okay, will try to get this done today, thanks already for the great work @michalk8!

cellrank/tl/_lineage.py
```diff
@@ -706,6 +706,44 @@ def copy(self, _="C") -> "Lineage":
...
         return self.copy()
 
+    def priming_degree(self, early_cells: Optional[np.ndarray] = None) -> np.ndarray:
+        """
+        TODO.
+
+        Parameters
+        ----------
+        early_cells
+            Indices or a mask marking early cells. If `None`, use all cells.
```
Marius1311, line 716: I don't agree with this - I think we need a way to specify the set of early cells on the cluster level. I know this makes it difficult to implement this into the Lineage class, which is why I think we should integrate it directly into the base estimator. Using clusters is the natural way to specify a set of early cells.

cellrank/tl/_lineage.py
```diff
@@ -706,6 +706,44 @@ def copy(self, _="C") -> "Lineage":
     def __copy__(self):
         return self.copy()
 
+    def priming_degree(self, early_cells: Optional[np.ndarray] = None) -> np.ndarray:
+        """
+        TODO.
```
Marius1311, line 711: Wait, why do we have this? I thought we went for something unified that can compute the lineage priming using either kl divergence or entropy?

cellrank/tl/estimators/_base_estimator.py
```diff
@@ -317,8 +316,7 @@ def compute_absorption_probabilities(
         Compute absorption probabilities of a Markov chain.
 
         For each cell, this computes the probability of it reaching any of the approximate recurrent classes defined
-        by :paramref:`{fs}`. This also computes the entropy over absorption probabilities, which is a measure of cell
-        plasticity, see [Setty19]_.
```
Marius1311, line 321: Totally fine to remove the sentence about priming degree here.

cellrank/tl/estimators/_base_estimator.py
```diff
@@ -555,6 +532,60 @@ def compute_absorption_probabilities(
...
+
+                - `'kl_divergence'` - TODO.
+                - `'entropy'` - TODO.
+
+        early_cells
+            Cell ids or a mask marking early cells. If `None`, use all cells. Only used when ``method='kl_divergence'``.
+
+        Returns
+        -------
+        The priming degree.
```
Marius1311, line 554: Inconsistent. Let's please agree on consistent naming: 
- the method is called `compute_lineage_priming`
- the method should return (and write) `lineage_priming`
- it should have modes `kl_divergence` and `entropy`
- accordingly, the corresponding lineage class methods should be `kl_divergence` and `entropy`, not `priming_degree` and `entropy`, I find this confusing.

Marius1311: So we can change to: "The method returns the degree of lineage priming."

cellrank/tl/estimators/_base_estimator.py
```diff
@@ -555,6 +532,60 @@ def compute_absorption_probabilities(
...
+                "Compute absorption probabilities first as `.compute_absorption_probabilities()`."
+            )
+        method = PrimingDegree(method)
+
+        if method == PrimingDegree.KL_DIVERERGENCE:
+            if early_cells is not None:
+                early_cells = np.asarray(early_cells)
+                if not np.issubdtype(early_cells.dtype, np.bool_):
+                    early_cells = np.isin(self.adata.obs_names, early_cells)
+            values = abs_probs.priming_degree(early_cells)
```
Marius1311, line 568: As I said above, I think this should be `abs_probs.kl_divergence`

cellrank/tl/estimators/_base_estimator.py
```diff
@@ -555,6 +532,60 @@ def compute_absorption_probabilities(
 
         self._write_absorption_probabilities(time=start, extra_msg=extra_msg)
 
+    def compute_lineage_priming(
+        self, method: str = "kl_divergence", early_cells: Optional[Sequence[str]] = None
+    ) -> pd.Series:
+        """
+        TODO.
```
Marius1311, line 539: Suggestion: 

"Compute the degree of lineage priming. 

This method computes how naive vs. committed each individual cell is. It returns a score where 0 stands for naive and 1 stands for committed. "

cellrank/tl/estimators/_base_estimator.py
```diff
@@ -555,6 +532,60 @@ def compute_absorption_probabilities(
...
+
+        Parameters
+        ----------
+        method
+            TODO. Valid options are:
```
Marius1311, line 544: Suggestion:

"The method used to compute the degree of lineage priming. Valid options are: 
- `kl_divergence`: as in [cite Stemnet], computes the kl divergence between the fate probabilities of a cell and the average fate probabilities. Computation of average fate probabilities can be restricted to a set of user-defined early cells, see [cite parameter]. The further from average a cell's fate probabilities, the more committed it is. 
- `entropy`, as in [palantir], computes entropy over a cell's fate probabilities. The lower a cell's entropy over fate probabilities, the more committed it is. "

Marius1311 (changes requested):
Apart from some naming changes I would suggest, this looks good. The only major comment I have is that I would really vote for being able to specify the set of early cells via a cluster key - this is much more convenient for the user, and really the natural way to do this I fee.

cellrank/tl/_lineage.py
michalk8: I'd implement it in both, since this really a property of a lineage and delegate.
I wanted to avoid inconsistencies between the 2, though it would be just as simple to do `early_cells=adata.obs['clusters']`.isin(['foo', 'bar])). However, there's also a different problem:
- 1. we'd either have to supply `cluster_key: str` to know from which clustering to take the clusters
- 2. or we'd have to save `cluster_key` when calling `compute_macrostates`, though this is used only for naming and I would advise against it (can also be `None`)
- 3. combination of both (also agains this)

I really don't want to introduce additional argument just for the cluster key, how about a compromise?:
- you can ALSO pass `{'cluster_key': ['foo', 'bar']}` and it will select only cells in `adata.obs['cluster_key'].isin(['foo', 'bar'])`

cellrank/tl/estimators/_base_estimator.py
michalk8: Ok, so I should remove `entropy` from `__overloaded_functions__` in Lineage, is that right?

michalk8:
Numba 0.53.0 seems to have broken many things, will have to take a look into it (i.e. whether we need to pin it down).

Marius1311:
Fantastic, thanks for finishing this off!

PR #516, opened 2021-03-09
Add a check for irreducibility

### LINKED ISSUE #515: Add a check for transition matrix irreducibility
- [x] Other?

We currently have a utility function which can decompose a (directed) transition matrix into recurrent and transient classes. We use this utility function in the base estimator class when we compute absorption probabilities to figure out whether or matrix is irreducible - however, that's an overkill. It's enough to check whether there are more than one strongly connected component, which runs in linear time. As soon as there's more than one such component, the matrix is reducible, no matter whether these components are recurrent or transient. 

So I propose to include a utility function for this which will be used in the base kernel class when we compute a transition matrix as an optional check for irreducibility. Additionally, I propose to use the new utility function when we compute absorption probabilities to check for irreducibility, as the old function is an overkill and takes much too long. We will still keep the old function (`_partition`), because it gives additional information which will be interesting for small-scale examples.

Marius1311:

## Title
Adds a utility function to check for irreducibility of a transition matrix and integrates this into kernel calsses

## Types of changes

## Description
This adds a utility function `_irreducible` to `cr.tl._utils` that checks whether a matrix is irreducible. The way we used to do this was computationally unnecessarily expensive because we also checked for each strongly connected component whether its recurrent or transient. 

## How has this been tested?
Via tests on the new utility function. Additionally, added some tests for `_connected`, `_symmetric` and `_partition` via 4 new test matrices in `_helpers.py`

## Closes
closes #515


cellrank/tl/_utils.py
```diff
@@ -769,6 +769,16 @@ def _connected(c: Union[spmatrix, np.ndarray]) -> bool:
...
 
 
+def _irreducible(d: Union[spmatrix, np.ndarray]) -> bool:
+    """Check whether the uirected graph encoded by d is irreducible."""
+
+    import networkx as nx
+
+    G = nx.DiGraph(d) if not isinstance(d, nx.DiGraph) else d
+
+    return len(list(nx.strongly_connected_components(G))) == 1
```
michalk8, line 779: Minor issue:
this will require to full evaluete the connected components (can be slow?). We just need to poll first 2 values to know
for sure whether it's irreducible or not. The function returns a generator, should be doable with `next` + `StopIteration` exception (I'd check whether it's a null graph as well first).

tests/_helpers.py
```diff
@@ -368,3 +368,406 @@ def _import_rpy2_mgcv() -> bool:
...
+    -------
+    :class:`numpy.ndarray`
+        Row-normalized transition matrix. This matrix is
+        - connected
+        - not irreducible (1 recurrent, 1 transient class)
+        - not reversible
+    """
+
+    p = np.array(
+        [
```
michalk8, line 423: Could you please reformat the array back to a block structure? It's really hard to read - you should you:
```python
# fmt: off
<the array>
# fmt: on
```
see [here](https://github.com/psf/black#the-black-code-style).

tests/_helpers.py
```diff
@@ -368,3 +368,406 @@ def _import_rpy2_mgcv() -> bool:
...
+            # 0.   1.   2.   3.   4.   5.   6.   7.   8.   9.   10.  11.  12.  13.
+            [0.0, 0.8, 0.2, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],  # 0
+            [0.2, 0.0, 0.6, 0.2, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],  # 1
+            [0.6, 0.2, 0.0, 0.2, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],  # 2
+            [
+                0.0,
```
michalk8, line 601: Also reformat this back please.

tests/_helpers.py
```diff
@@ -368,3 +368,406 @@ def _import_rpy2_mgcv() -> bool:
 gamr_skip = pytest.mark.skipif(
     _import_rpy2_mgcv(), reason="Cannot import `rpy2` or R's `mgcv` package."
 )
+
+
+def test_matrix_1() -> np.ndarray:
```
michalk8, line 373: Should not be in `_helpers.py`, should be a `@pytest.fixture(scope='session')` in `conftest.py`.
It will get instantiated only once (i.e. more efficient) - useful since you don't write to it.

tests/_helpers.py
```diff
@@ -368,3 +368,406 @@ def _import_rpy2_mgcv() -> bool:
...
+            [0.0, 0.0, 0.0, 0.0, 0.05, 0.05, 0.0, 0.0, 0.0, 0.0, 0.7, 0.2],  # 9
+            [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.2, 0.0, 0.8],  # 10
+            [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.8, 0.2, 0.0],  # 11
+        ]
+    )
+
+    return p
+
+
+def test_matrix_2() -> np.ndarray:
```
michalk8, line 408: Should be a fixture.

tests/_helpers.py
```diff
@@ -368,3 +368,406 @@ def _import_rpy2_mgcv() -> bool:
...
+                0.3,
+                0.7,
+            ],  # 13
+        ]
+    )
+
+    return p
+
+
+def test_matrix_3() -> np.ndarray:
```
michalk8, line 580: Should be a fixture.

tests/_helpers.py
```diff
@@ -368,3 +368,406 @@ def _import_rpy2_mgcv() -> bool:
...
+                0.3,
+                0.7,
+            ],  # 13
+        ]
+    )
+
+    return p
+
+
+def test_matrix_4() -> np.ndarray:
```
michalk8, line 752: Should be a fixture.

tests/test_utils.py
```diff
@@ -1,5 +1,13 @@
 import pytest
-from _helpers import create_model, assert_array_nan_equal, jax_not_installed_skip
+from _helpers import (
+    create_model,
```
michalk8, line 3: The `test_matrix_{1,2,3,4}` can be removed, once they become fixtures.

cellrank/tl/_utils.py
```diff
@@ -769,6 +769,16 @@ def _connected(c: Union[spmatrix, np.ndarray]) -> bool:
     return nx.is_connected(G)
 
 
+def _irreducible(d: Union[spmatrix, np.ndarray]) -> bool:
```
michalk8, line 772: Since you're always logging after calling this function (at least in the function in this PR), consider
moving the log calls inside this function.
If there's a call where the callee afterwards does not log, please ignore this comment.

cellrank/tl/kernels/_base_kernel.py
```diff
@@ -646,6 +648,13 @@ def _compute_transition_matrix(
             )
             matrix[problematic_indices, problematic_indices] = 1.0
 
+        if check_irreducibility:
```
michalk8, line 651: For consistency, I'd either name it as in base_estimator (`check_irred`), or vice-versa.

cellrank/tl/kernels/_velocity_kernel.py
```diff
@@ -180,6 +181,8 @@ def compute_transition_matrix(
             Number of bootstrap samples when ``mode={m.MONTE_CARLO.s!r}``.
         seed
             Set the seed for random state when the method requires ``n_samples``.
+        check_irreducibility
+            Optional check for irreducibility of the final transition matrix.
```
michalk8, line 185: Same inconsistency as raised above.

cellrank/tl/kernels/_palantir_kernel.py
```diff
@@ -97,6 +100,8 @@ def compute_transition_matrix(
             This is done to ensure that the graph remains connected.
         density_normalize
             Whether or not to use the underlying KNN graph for density normalization.
+        check_irreducibility
```
michalk8, line 103: Same naming inconsistency as above.

michalk8 (changes requested):

cellrank/tl/_utils.py
Marius1311: Hi Mike, finding connected components runs in linear time and is thus very fast. If you have something in mind already then please feel free to add, otherwise, I would leave this as it is. What's a null graph?

Marius1311:
Hi @michalk8, thanks for the helpful comments! Will fix them and update.

cellrank/tl/_utils.py
Marius1311: That's a good point, thanks.

tests/_helpers.py
Marius1311: done.

tests/_helpers.py
Marius1311: done.

tests/_helpers.py
Marius1311: done.

tests/_helpers.py
Marius1311: done.

tests/_helpers.py
Marius1311: done.

cellrank/tl/_utils.py
Marius1311: done.

cellrank/tl/kernels/_base_kernel.py
Marius1311: I went for `check_irreducibility` and changed this in the `base_estimator`.

cellrank/tl/kernels/_palantir_kernel.py
Marius1311: done.

Marius1311:
Hi @michalk8, all done, apart from your top comment. I don't think that's 100% necessary since this is very efficient (linear in cell number, 3sec for 12k cells), but if you have a quick fix in mind, feel free to go ahead. Otherwise, I will merge as soon as the tests pass.

michalk8:
I've fixed the tests + use the iterator in `_irreducible`. Merging.

Marius1311:
Fantastic, thanks!

PR #520, opened 2021-03-11
Add compute_projection

### LINKED ISSUE #519: Project transition matrix in the embedding
- [x] Other?

Similarly as scvelo does it.

michalk8:

## Types of changes

## Description

Add new method for `KernelExpression` - `comput_projection(basis: str = 'umap', key_added: Optional[str] = None, copy: bool = False)`.
Here's how it looks like when running:
```python
import cellrank as cr
import scvelo as scv

adata = cr.datasets.pancreas_preprocessed()
T = cr.tl.kernels.PalantirKernel(adata, backward=False).compute_transition_matrix()
T.compute_projection()
scv.pl.velocity_embedding(adata, arrow_length=4, arrow_size=4, dpi=180, vkey='T_fwd', basis='umap')
```
index
pinging @Marius1311 : any idea why the directions are reversed? For pure velocity kernel, they are fine.

Also @Marius1311 , can you shed some light on `retain_scale` argument, used [here](https://github.com/theislab/scvelo/blob/master/scvelo/tools/velocity_embedding.py#L165)?

## How has this been tested?

Not yet.

## Closes

closes #519 

TODOs:
- [ ] handle dense matrices
- [ ] finalize the API (I've always do `autoscale=True`, maybe expose?)
- [ ] write params to .uns? (don't think it's really needed)
- [ ] tests


Marius1311:
Uh, I think for the `retain_scale` argument, @WeilerP may be the right person to ask. To put this in context for you @WeilerP, we're working on visualising arbitary transition matrices via arrows in the embedding - for the image @michalk8 posted above, we never used RNA velocity information.

Marius1311:
@michalk8, why did you make the decision to integrate this into the kernel class, rather than the estimator class?

Marius1311:
Re the directions, please post the pseudotime used here, otherwise, I can't judge this.

Marius1311:
Well done btw!

Marius1311:
I shall review this after my PR today.

michalk8:
> @michalk8, why did you make the decision to integrate this into the kernel class, rather than the estimator class?

Because it makes more sense. We can always wrap in the estimator if needed (vice-versa is not doable)

Marius1311:
> > @michalk8, why did you make the decision to integrate this into the kernel class, rather than the estimator class?
> 
> Because it makes more sense. We can always wrap in the estimator if needed (vice-versa is not doable)

Okay, agreed - it's basically only a visualisation of the transition matrix, so fits into a kernel.

cellrank/tl/kernels/_base_kernel.py
```diff
@@ -223,6 +227,85 @@ def kernels(self) -> List["Kernel"]:
...
+        Compute a projection of the transition matrix in the embedding.
+
+        The projected matrix can be then visualized as::
+
+            scvelo.pl.velocity_embedding(adata, vkey='T_fwd', basis='umap')
+
+        Parameters
+        ----------
+        basis
+            Basis for which to compute the projection.
```
Marius1311, line 246: Basis for which to compute the projection, accessed via :paramref:`adata` ``.obsm['X_{basis}']``

Marius1311 (commented):
Sorry need to leave, will get back to this tomorrow.

WeilerP:
> Uh, I think for the `retain_scale` argument, @WeilerP may be the right person to ask. To put this in context for you @WeilerP, we're working on visualising arbitary transition matrices via arrows in the embedding - for the image @michalk8 posted above, we never used RNA velocity information.

Sorry for answering only now. Using `retain_scale=True` will reflect the length (difference) of velocity vectors in the higher dimensional space in the lower dimensional one. I guess the name was chose since the scale (of norms) is the same in both low and high dimensional space (or not).

Marius1311:
ok, thanks @WeilerP!

Marius1311:
Okay, here's why I think this currently is not working as expected. In Fig. 1 below, consider one reference cells and its neighbors in the KNN graph as well as its pruned neighbors in the Palantir kernel transition matrix, where some neighbors have been removed. Pseudotime on the right. In Fig. 2, I zoom in onto the neighbors in the pruned transition matrix. The arrows have been computed using the palantir kernel transition matrix. Pseudotime (in the middle) has been centered around the reference cell, you can see that most cells lie in the pseudotime-future. However, once we pruned the KNN graph, all that matters for transition probabilities is connectivities, and since we kept the 3 nearest neighbors, even though they may be in the pseudotime past, the arrow may very often point "backwards. 

## Fig. 1
image

## Fig. 2

Marius1311:
I suspect that this is a problem of the hard thresholding scheme and that the soft thresholding scheme, maybe with some slight adjustments, will resolve this. From what I can see, this is not a problem with your re-implementation of the projection method.

Marius1311:
Fig. 1 below shows what I get when I compute the transition matrix (based on the same pseudotime) using the `soft` scheme via
```
pk = PseudotimeKernel(adata)
pk.compute_transition_matrix(threshold_scheme='soft', nu=0.5, b=40, percentile=100)
pk.compute_projection()
scv.pl.velocity_embedding(adata, arrow_length=4, arrow_size=4, dpi=150, vkey='T_fwd')
```

## Fig. 1
image

I conclude that the backwards-arrow problem we're observing above is really a problem of our current implementation of the `hard` thresholding scheme and not of the embedding projection. Hence, I recommend we merge this PR.

michalk8:
Ok, agreed, let's continue the discussion in #530

PR #536, opened 2021-03-19
Fix dist -> conn, update docstrings, remove copy

### LINKED ISSUE #530: Description of `k` in `hard` `PseudotimeKernel` is wrong
I basically got this wrong when I wrote down the doscring, currently, it says:
"k
    Number of neighbors to keep for each node, regardless of pseudotime.
    This is done to ensure that the graph remains connected. Only used when `threshold_scheme='hard'`."

That is **not** correct. However, since we have have to change this scheme a bit anyways, let's include the docstring change into the next PR.

michalk8:

## Types of changes

## Closes

closes #530 

TODOs:
- [ ]  use PT to prune past edges, not connectivities
- [ ] change adaptive `k`


Marius1311:
We need to discuss how we want to get the density dependency right.

Marius1311:
Hi Mike, I find the random walk plotting very insightful for the `hard` pseudotime kernel. Check below for a pseudotime (Fig. 1) that was used to define a default-parameter `hard` pseudotime kernel. See Fig. 2 for the projected transition matrix. Arrows point backwards, as before. However, observe how, when we initialize 10 random walks in the intermediate Fev+ cluster, they all go into the correct direction! This makes me wonder - could there be something wrong with the embedding projection, or, more specifically, could there be any problem that arises in this projection when we prune neighbors, i.e. does the projection assume something about the number of neighbors?

## Fig. 1
image

## Fig. 2
image

## Fig. 3
image

michalk8:
Really strange, based on what I've tried just now, when setting the past connections to 1e-12, I get dramatically different results:
1e-12
Here's the function I'm using
```python
def fn1(
        cell_pseudotime: float,
        neigh_pseudotime: np.ndarray,
        neigh_dist: np.ndarray,
        n_neighs: int = 30,
        k: int = 3,
    ):
    k_thresh = max(0, min(30, int(np.floor(n_neighs / k)) - 1))

    # below code does not work with argpartition
    ixs = np.flip(np.argsort(neigh_dist))
    close_ixs, far_ixs = ixs[:k_thresh], ixs[k_thresh:]

    mask_keep = cell_pseudotime <= neigh_pseudotime[far_ixs]
    far_ixs_keep = far_ixs[mask_keep]

    biased_dist = np.zeros_like(neigh_dist) + 1e-12
    biased_dist[close_ixs] = neigh_dist[close_ixs]
    biased_dist[far_ixs_keep] = neigh_dist[far_ixs_keep]

    return biased_dist
```

Marius1311:
Hi Mike, I decided to go for very minimal changes: I went from `k` to `fract_to_keep` which is the fraction of neighbors to keep where neighbors are sorted by connectivities. That means the method behaviour actually stays the same, we're still pruning on connectivities and not on the pseudotime. I didn't change that because I really have no intuition for what's best here. What mattered most from my point of view is that the docstring correctly describes what's going on, and this has been fixed - I thikn we can merge this. 

If we want to, we can always go back to the question of pruning later and experiment whether it's better to prune on connectivities or pseudotimes.

Marius1311:
@michalk8, can I assume that in your function above, what you call distances are really connectivites? This really confuses me in the interpretation.

michalk8:
> @michalk8, can I assume that in your function above, what you call distances are really connectivites? This really confuses me in the interpretation.

Yes, sorry, I copied it from earlier.

Marius1311:
Okay, could you check whether something's going wrong the way we select far and close cells, and the way we prune, i.e. slicing etc.?

michalk8:
> Okay, could you check whether something's going wrong the way we select far and close cells, and the way we prune, i.e. slicing etc.?

I did, it's ok. If I use the `1e-12` vs not using it, what happens when not using it is that we prune some low. prob. connectivities:
```python
array([[1.88810865e-13, 1.70982877e-13, 2.42167932e-13, 1.77058177e-13,
        2.22136713e-13, 1.64979464e-13, 1.91707062e-13, 2.28112563e-13,
        1.62352468e-13, 2.11157980e-13]])
```
In total, it's just `1e-12` of the mass, but the shift in mean is:
```python
0.029411764705882346 -> 0.04166666666666666
```

Marius1311:
Ok, on the level of the random walks, everything makes sense to me. In the figures below, I'm varying `fract_to_keep`, my expectation being that if I keep a larger fraction of neighbors, I will see more variance, i.e. more random walks going against the direction of pseudotime. That's exactly what these plots tell me. 

## Fig. 1: `fract_to_keep=0.1`
image

## Fig. 2: `fract_to_keep=0.3`
image

## Fig. 3: `fract_to_keep=0.7`
image

Marius1311:
Equally, the kernel behaves as expected on the level of the spectrum, again, varying `fract_to_keep` and showing the top eigenvalues. As I increase `fract_to_keep`, I would expect the spectrum to "smooth out", i.e. the gaps become smaller, the spectrum approaches that of a ConnectivityKernel. 

## Fig. 1: `fract_to_keep=0.1`
image

## Fig. 2: `fract_to_keep=0.3`
image

## Fig. 3: `fract_to_keep=0.7`
image

Marius1311:
So far, I feel like this is a problem with the embedding projection.

michalk8:
Agreed, the plots make sense, I think we can merge this (I've fixed the tests + changed `fract_to_keep` to `frac_to_keep` [frac is more common abbreviation]).

michalk8:
As for the projection, not sure how best to fix it (it works when nor removing the connections using VIA-like scheme + it also works for VelocityKernel).

Marius1311:
Yes - let's merge this once the test pass. It's a problem with the embedding projection that we should keep in mind.

PR #595, opened 2021-05-27
Exp time kernel

### LINKED ISSUE #551: Real-time integration
- [ ] Other?

This is about including a kernel that computes a directed transition matrix base on real-time information. Some options that come to my mind:
- WaddingtonOT [WOT] (pairwise unbalanced optimal transport among time-points), see https://www.cell.com/cell/pdf/S0092-8674(19)30039-X.pdf
- global Waddington OT [gWOT] (same as WOT, but OT problem solved simultaneously for all time points rather than pairwise), see https://github.com/zsteve/gWOT
- Dana's harmony (heuristic that uses mutual nearested neighbors in conjunction with regression to estimate the existence and strength of inter-timepoint links), see https://www.nature.com/articles/s41586-019-1127-1 or https://github.com/dpeerlab/Harmony

These are just three, there are probably more out there. Architecture-wise, it would probably make most sense to include this via the `external` interface.

michalk8:

## Types of changes

## Description

Add Waddington OT kernel, see related issue https://github.com/broadinstitute/wot/issues/91. I've also added option to use existing obs features (PCA, imputed expression) to compute the cost matrices.
Also adds `ExpTimeKernel` class which can be later used to derive kernels with experimental time.

@Marius1311 I've decided against adding anything to the block diagonal, don't think it's a nice solution - user can always weight with full `ConnectivitiesKernel` and afaik, the GPCCA stability issues were gone (I've used small weights like 0.05 or so for connectivities). Maybe you add some nice note in `.compute_transition_matrix`, below `Returns` as
```
Notes
-----
...
```

## How has this been tested?

I've added a few.

## Closes

closes #551

### DIFF
--- a/cellrank/external/kernels/__init__.py
+++ b/cellrank/external/kernels/__init__.py
@@ -1,2 +1,2 @@
 from cellrank.external.kernels._wot_kernel import WOTKernel
-from cellrank.external.kernels._statot_kernel import OTKernel
+from cellrank.external.kernels._statot_kernel import StationaryOTKernel
--- a/cellrank/external/kernels/__init__.py
+++ b/cellrank/external/kernels/__init__.py
@@ -1,1 +1,2 @@
+from cellrank.external.kernels._wot_kernel import WOTKernel
 from cellrank.external.kernels._statot_kernel import OTKernel
--- a/cellrank/external/kernels/_statot_kernel.py
+++ b/cellrank/external/kernels/_statot_kernel.py
@@ -23,7 +23,7 @@ except ImportError as e:
 
 
 @d.dedent
-class OTKernel(OTKernel_, error=_error):
+class StationaryOTKernel(OTKernel_, error=_error):
     """
     Stationary optimal transport kernel from [Zhang21]_.
 
@@ -76,7 +76,7 @@ class OTKernel(OTKernel_, error=_error):
         C: Optional[np.ndarray] = None,
         verbose: bool = False,
         **kwargs: Any,
-    ) -> "OTKernel":
+    ) -> "StationaryOTKernel":
         """
         Compute transition matrix using stationary OT [Zhang21]_.
 
@@ -113,7 +113,7 @@ class OTKernel(OTKernel_, error=_error):
 
         Returns
         -------
-        :class:`cellrank.external.kernels.OTKernel`
+        :class:`cellrank.external.kernels.StationaryOTKernel`
             Makes :attr:`transition_matrix` available.
         """
         if method not in ("ent", "quad", "unbal"):
--- a/cellrank/external/kernels/_wot_kernel.py
+++ b/cellrank/external/kernels/_wot_kernel.py
@@ -49,6 +49,11 @@ class WOTKernel(Kernel, error=_error):
     %(cond_num)s
     """  # noqa: E501
 
+    __import_error_message__ = (
+        "Unable to import the kernel. Please install `wot` first as "
+        "`pip install git+https://github.com/broadinstitute/wot`."
+    )
+
     def __init__(
         self,
         adata: AnnData,
@@ -70,6 +75,7 @@ class WOTKernel(Kernel, error=_error):
         )
         self.adata.obs[self._time_key] = self.experimental_time
         self._growth_rates = None
+        self._tmats = None
 
     def compute_transition_matrix(
         self,
@@ -88,8 +94,7 @@ class WOTKernel(Kernel, error=_error):
         cost_matrices
             Cost matrices for each consecutive time pair.
             If a :class:`str`, it specifies a key in :attr:`adata` ``.layers``: or :attr:`adata` ``.obsm``
-            containing cell features that are used to compute cost matrices.
-            If `None`, use `WOT`'s default.
+            containing cell features that are used to compute cost matrices. If `None`, use `WOT`'s default.
         solver
             Which solver to use.
         growth_rate_key
@@ -164,7 +169,7 @@ class WOTKernel(Kernel, error=_error):
             **kwargs,
         )
 
-        tmats: Dict[Tuple[float, float], AnnData] = {}
+        self._tmats: Dict[Tuple[float, float], AnnData] = {}
         start = logg.info(
             f"Computing transport maps for `{len(cost_matrices)}` time pairs"
         )
@@ -177,12 +182,11 @@ class WOTKernel(Kernel, error=_error):
                 raise ValueError(
                     f"Encountered `{nans}` NaN values for time pair `{tpair}`."
                 )
-            tmat.X = _normalize(tmat.X)
-            tmats[tpair] = tmat
+            self._tmats[tpair] = tmat
 
         logg.info("    Finish", time=start)
 
-        return tmats
+        return self._tmats
 
     def _restich_tmats(self, tmaps: Mapping[Union[float, float], AnnData]) -> spmatrix:
         blocks = [[None] * (len(tmaps) + 1) for _ in range(len(tmaps) + 1)]
@@ -190,7 +194,7 @@ class WOTKernel(Kernel, error=_error):
         obs_names, growth_rates = [], []
 
         for i, tmap in enumerate(tmaps.values()):
-            blocks[i][i + 1] = tmap.X
+            blocks[i][i + 1] = _normalize(tmap.X)
             nrows += tmap.n_obs
             ncols += tmap.n_vars
             obs_names.extend(tmap.obs_names)
@@ -198,8 +202,8 @@ class WOTKernel(Kernel, error=_error):
         obs_names.extend(tmap.var_names)
 
         n = self.adata.n_obs - nrows
-        blocks[-1][-1] = spdiags([1] * n, 0, n, n)
-        # prevent from disappearing
+        blocks[-1][-1] = np.ones((n, n)) / float(n)
+        # prevent block from disappearing
         n = blocks[0][1].shape[0]
         blocks[0][0] = spdiags([], 0, n, n)
 
@@ -297,6 +301,11 @@ class WOTKernel(Kernel, error=_error):
     def growth_rates(self) -> Optional[pd.DataFrame]:
         """Estimated cell growth rates for each growth rate iteration."""
         return self._growth_rates
+
+    @property
+    def transition_maps(self) -> Optional[Dict[Tuple[float, float], AnnData]]:
+        """Transition maps for each consecutive time pair."""
+        return self._tmats
 
     def __invert__(self) -> "WOTKernel":
         super().__invert__()
--- a/cellrank/external/kernels/_wot_kernel.py
+++ b/cellrank/external/kernels/_wot_kernel.py
@@ -95,7 +95,7 @@ class WOTKernel(Kernel, error=_error):
         growth_rate_key
             Key in :attr:`adata` ``.obs`` where initial cell growth rates are stored.
         kwargs
-            Keyword arguments for OT configuration.
+            Additional keyword arguments for OT configuration.
 
         Returns
         -------
@@ -110,10 +110,10 @@ class WOTKernel(Kernel, error=_error):
         kwargs.setdefault("epsilon", 0.05)
         kwargs.setdefault("lambda1", 1)
         kwargs.setdefault("lamdda2", 50)
-        kwargs["growth_iters"] = max(kwargs.get("growth_iters"), 1)
+        kwargs["growth_iters"] = max(kwargs.get("growth_iters", 1), 1)
 
         start = logg.info(
-            "Computing transition matrix using Waddington Optimal Transport"
+            "Computing transition matrix using Waddington optimal transport"
         )
 
         cost_matrices, cmat_param = self._generate_cost_matrices(cost_matrices)
@@ -199,6 +199,7 @@ class WOTKernel(Kernel, error=_error):
 
         n = self.adata.n_obs - nrows
         blocks[-1][-1] = spdiags([1] * n, 0, n, n)
+        # prevent from disappearing
         n = blocks[0][1].shape[0]
         blocks[0][0] = spdiags([], 0, n, n)
 
@@ -296,3 +297,9 @@ class WOTKernel(Kernel, error=_error):
     def growth_rates(self) -> Optional[pd.DataFrame]:
         """Estimated cell growth rates for each growth rate iteration."""
         return self._growth_rates
+
+    def __invert__(self) -> "WOTKernel":
+        super().__invert__()
+        # because WOT reads from `adata`
+        self.adata.obs[self._time_key] = self.experimental_time
+        return self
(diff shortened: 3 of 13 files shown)


Marius1311:
Hi Mike, first comment, when trying to import with `otk = WOTKernel(adata)` (I don't have WOT installed yet), I get
```python
ModuleNotFoundError                       Traceback (most recent call last)
~/Projects/cellrank/cellrank/external/kernels/_wot_kernel.py in <module>
     17 try:
---> 18     import wot
     19 

ModuleNotFoundError: No module named 'wot'

The above exception was the direct cause of the following exception:

ImportError                               Traceback (most recent call last)
<ipython-input-20-d38dddb5f66c> in <module>
----> 1 otk = WOTKernel(adata)

~/Projects/cellrank/cellrank/external/kernels/_wot_kernel.py in __init__(self, adata, backward, time_key, compute_cond_num)
     57         compute_cond_num: bool = False,
     58     ):
---> 59         super().__init__(
     60             adata,
     61             backward=backward,

~/Projects/cellrank/cellrank/external/_error_mixin.py in __init__(self, *args, **kwargs)
     24 
     25     def __init__(self, *args: Any, **kwargs: Any):
---> 26         raise ImportError(self.__import_error_message__) from self.__error__
     27 
     28     def __init_subclass__(cls, error: Optional[Exception] = None, **kwargs: Any):

ImportError: Unable to import the kernel.
```

This is of course expected as I don't have the package installed - however, could we print a message, telling me to install it from pip?

michalk8:
Hey @Marius1311 , done. I've also exposed `.transition_maps` + changed the last time pair block not to sparse diagonal, but to normalized all 1s.

Marius1311:
I installed WOT and initalized the kernel, as above. I get the following warning: `WARNING: Time categories are not ordered. Using default order`. What is default order in this context?

Marius1311:
Next point (I'm just going through as I experiment, please excuse the absence of structure here): I print my initialized kernel object, not transition matrix computed yet: `print(otk)` gives `<WOTKernel[cost_matrices=default, solver=duality_gap, growth_rate_key=None, epsilon=0.05, lambda1=1, lamdda2=50, growth_iters=1]>`

I'm just wondering why so many of these are object level properties - I suppose they are specified when I compute a transition matrix?

michalk8:
> I installed WOT and initalized the kernel, as above. I get the following warning: WARNING: Time categories are not ordered. Using default order. What is default order in this context?

It's in ascending order, changed the message.

> I'm just wondering why so many of these are object level properties - I suppose they are specified when I compute a transition matrix?

There's even more if you supply kwargs, see all options [here](https://github.com/broadinstitute/wot/blob/master/wot/ot/ot_model.py#L85). I don't think this is problematic though.

Marius1311:
I updated my post from above - I don't think it's problematic either, I just wonder why we handle it differently here than in other kernels.

Marius1311:
Next point: I think `otk.compute_transition_matrix()` should not absorb `epsilon, lambda1, lamdda2, growth_iters` into the `kwargs` - these are really important arguments and the user should think about them when running the method. Can you please add these to the docstring explicitly, and also emhasize that the user should check out the original documentation (provide a link) before running the function?

michalk8:
> I print my initialized kernel object, not transition matrix computed yet: print(otk) gives <WOTKernel[cost_matrices=default, solver=duality_gap, growth_rate_key=None, epsilon=0.05, lambda1=1, lamdda2=50, growth_iters=1]>

Can't reproduce, only after `.compute_transition_matrix` you see all the params.

> should not absorb epsilon, lambda1, lamdda2, growth_iters into the kwargs

Ok.

> and also emhasize that the user should check out the original documentation (provide a link) before running the function?

Can't do, afaik there are not online docs. https://broadinstitute.github.io/wot/ only has tutorials + cmd line interface + additional stuff.

Marius1311:
> > I print my initialized kernel object, not transition matrix computed yet: print(otk) gives <WOTKernel[cost_matrices=default, solver=duality_gap, growth_rate_key=None, epsilon=0.05, lambda1=1, lamdda2=50, growth_iters=1]>
> 
> Can't reproduce, only after `.compute_transition_matrix` you see all the params.
> 

You are right, I must have messed it up somehow. All good here. 

> > should not absorb epsilon, lambda1, lamdda2, growth_iters into the kwargs
> 
> Ok.
> 
> > and also emhasize that the user should check out the original documentation (provide a link) before running the function?
> 
> Can't do, afaik there are no online docs. https://broadinstitute.github.io/wot/ only has tutorials + cmd line interface + additional stuff.

It's fine to link to that page, it's all explained in the tutorials in length, that's where I got the info from.

Marius1311:
Is `otk.compute_transition_matrix()` parallelised across time-point pairs?

Marius1311:
Some more open points: 
- We need an example for the usage in the docs. It's currently not clear to me how that data is expected to be pre-processed before feeding it into the kernel
- We should brainstorm convenient ways to initialize the `growt_rate`, essentially folllowing the second code cell in https://nbviewer.jupyter.org/github/broadinstitute/wot/blob/master/notebooks/Notebook-2-compute-transport-maps.ipynb

Marius1311:
Another point: can we write the final growth rates (last round of iteration) back to adata.obs please? I think that's an important quantity.

michalk8:
> Is otk.compute_transition_matrix() parallelised across time-point pairs?

No, and I don't think it should be - the OT solver is already consuming all cores.

> We need an example for the usage in the docs. It's currently not clear to me how that data is expected to be pre-processed before feeding it into the kernel

Ok, can you please add this? I also don't know which pre-processing should be used (would be best), just used the imputed data from scVelo (standard preprocessing).

> Another point: can we write the final growth rates (last round of iteration) back to adata.obs please? I think that's an important quantity.

Ok, done.

> We should brainstorm convenient ways to initialize the growt_rate, essentially folllowing the second code cell in https://nbviewer.jupyter.org/github/broadinstitute/wot/blob/master/notebooks/Notebook-2-compute-transport-maps.ipynb

How do you propose to do that? We don't have neither prolif./apop. info. 2 options I have in mind:
1. optionally pass these 2 cols to `__init__` and compute the BD process
2. pass gene sets as in notebook 1, cell 4 to compute the scores and get prolif./apop. info from that

1. is more general/flexible, 2. is more convenient.

Marius1311:
Hi Mike! 
- parallelisation: okay, got it. 
- workflow: okay, will add. Please add a key like `data_key` or so, where one can specify which layer/`.X` to use
- actually, for the PCA, do we use the precomputed one now? i.e. not pairwise PCAs? Since this is not default in WOT, should inform/warn the user about this and potentially give a choice. 
- growth rates: I would go for option 1, i.e. read cell-level proliferation/apoptosis scores from `.obs`. There is already a scanpy function for computing scores from gene sets, so we shouldn't reinvent the wheel. Can you check whether their gene lists are available somewhere in the github? 

Thanks! We're getting there.

michalk8:
> workflow: okay, will add. Please add a key like data_key or so, where one can specify which layer/.X to use

This is already done: `cost_matrices` can be a key in `.layers` or even `.obsm`.

 > actually, for the PCA, do we use the precomputed one now? i.e. not pairwise PCAs? Since this is not default in WOT, should inform/warn the user about this and potentially give a choice.

We're using WOT's default now (pairwise PCA).

> growth rates: I would go for option 1, i.e. read cell-level proliferation/apoptosis scores from .obs. There is already a scanpy function for computing scores from gene sets, so we shouldn't reinvent the wheel. Can you check whether their gene lists are available somewhere in the github?

Ok, agreed, will add this.

michalk8:
> growth rates: I would go for option 1, i.e. read cell-level proliferation/apoptosis scores from .obs. There is already a scanpy function for computing scores from gene sets, so we shouldn't reinvent the wheel. Can you check whether their gene lists are available somewhere in the github?

Done as `.compute_initial_growth_rates`. Please check the docs and see if you'd improve anything.
Currently envisioned workflow:
```python
# preprocessing

ok = cr.external.kernels.WOT(adata, time_key=...)
ok.compute_initial_growth_rates("prolif_key", "apop_key", ..., key_added="initial_growth_rates")  # optional
ok.compute_transition_matrix(..., growth_rate_key="initial_growth_rates")

**Comment by Marius1311** [comment]
Looks great, thanks @michalk8. Questions:
- PCA: is there an option to also compute a combined PCA representation? Or only pairwise? Also, for the pairwise PCA's, do you just use `.X` to compute them?
- What's the relationship between the `cost_matrices` and PCA? If I provide a cost matrix, is there no need to compute pairwise PCAs?
- Did you check whether you can find their gene sets somewhere in the repo?
- On my part, I will go over the docs, add the workflow example, and test this again.

**Comment by michalk8** [comment]
> Did you check whether you can find their gene sets somewhere in the repo?

Yes, they should be present, but I would not include them, because
```
The file data/gene_sets.gmx contains gene sets relevant to our reprogramming database, as explained in Schiebinger et al 2019. For a different application, you will need to use different gene sets.
```

> What's the relationship between the cost_matrices and PCA? If I provide a cost matrix, is there no need to compute pairwise PCAs?

If `cost_matrices` is None (default) the behavior is as in WOT (compute PCA only on time pair subset + euc. distance).
If `cost_matrices` is a string, find the representation in `.layers` ('X' for `.X`) or `.obsm` (in this order) and compute the euc distances.
If `cost_matrices` is a dict, the cost matrices should be precomputed for every time pair.

> PCA: is there an option to also compute a combined PCA representation? Or only pairwise? Also, for the pairwise PCA's, do you just use .X to compute them?

As mentioned above, you can use any repr. For pairwise, it's always on X, as in WOT.

> On my part, I will go over the docs, add the workflow example, and test this again.

Thanks.

Marius1311:
Thanks for clarifying! 
- gene sets: I think their comments relate mostly to the other gene sets they have in the file - I suspect they have gene sets in there that relate to their Fig. 2E, which defines e.g. "Neural identity", which of course is specific to this dataset. The signature we are interested in, i.e. proliferation and apoptosis, should be more general and should only depend on the organism, i.e. mouse, human, rat, etc. So I'm not saying we should hardcode them, it's more for me so I can use these in my own experiments. 
- cost matrices: got it, I like your implementation.

New question: are time points actually re-ordered in AnnData? Or how do you make sure the final transition matrix has the same  row and column ordering as the AnnData object?

Marius1311:
Also, I think we should add a tutorial for this, maybe in combination with StationaryOT. I can take care of this.

Marius1311:
Is scanpy used under the hood to compute PCA?

michalk8:
> New question: are time points actually re-ordered in AnnData? Or how do you make sure the final transition matrix has the same row and column ordering as the AnnData object?

The transition matrix has the same order of cells as the original one.

> Is scanpy used under the hood to compute PCA?

No.

Marius1311:
> > New question: are time points actually re-ordered in AnnData? Or how do you make sure the final transition matrix has the same row and column ordering as the AnnData object?
> 
> The transition matrix has the same order of cells as the original one.
So will the row/column ordering be the same as in `adata.obs`?

> 
> > Is scanpy used under the hood to compute PCA?
> 
> No.
What do you use? Is highly variable gene annotation taken into account?

michalk8:
> So will the row/column ordering be the same as in adata.obs?

Yes.

> What do you use? Is highly variable gene annotation taken into account?

No. Can add option to subset the adata before to use only HVGs only.

Marius1311:
Re HVGs, that would be good - I would have a parameter `use_highly_variable_genes: bool = True` that searches `adata.var` for a field called `highly_variable`. If it's present and it contains bool annotations, these are used to select genes, i.e. internal subsetting. If it's not present, a warning should be issued and all genes should be used.

Marius1311:
We should also rename that `OTKernel` - That should really be called `StationaryOTKernel`.

Marius1311:
For the record, I used the `WOTKernel` on the lung data and plotted the transition matrix, once without cell-ordering (Fig. 1) adn once with cell-ordering by time-point (Fig. 2) I see the expected patter, i.e. radomess in. Fig. 1, off-diagonal blocks in Fig. 2. Why did you decide not to decide self-loops only (diagonal only) for the last time-point?

## Fig. 1
Unordered: 
image

## Fig. 2
Ordered by time-point: 
image

Marius1311:
Note to self: we should add to the docstring how we assign transition probs to the last time-point.

Marius1311:
Random walks are currently very "jumpy": 
image

Could be a consequence of the way in which we assign transition probs to the last time-point - we will just get random jumps in the last time point, all over the UMAP. Probably not ideal.

michalk8:
> Why did you decide not to decide self-loops only (diagonal only) for the last time-point?

Because I can't justify using only diagonal.

> Note to self: we should add to the docstring how we assign transition probs to the last time-point.

Agreed, once we converge on solution, could you please update it?

Marius1311:
> > Why did you decide not to decide self-loops only (diagonal only) for the last time-point?
> 
> Because I can't justify using only diagonal.

I get that point - in case we know nothing, you say let's use the maximum entropy distribution here which would be uniform. However, taking into account computational considerations, I would go for sparse diagonal because it's 
1. sparse, i.e. takes less memory
2. terminates random walks. Otherwise, we essentially can't use that plotting function because you will get jumps all over the umap. 

> 
> > Note to self: we should add to the docstring how we assign transition probs to the last time-point.
> 
> Agreed, once we converge on solution, could you please update it?

Will do!

michalk8:
> Since there is no obvious choice here, how about we leave it to the user? t_max_map: str = 'diagonal' (or uniform). Would be convenient for me to check random walks.

Done as `uniform: bool = False`. I've also changed the cmat messages from debug -> info.

Marius1311:
Fantastic, thanks!

Marius1311:
I found the gene sets to compute the growth rates. See Fig. 1 for the proliferation and apoptosis scores and Fig. 2 for the combined computed growth rate. This matches annotations well, see Fig. 3, note the proliferation cluster gets higher growth rate. 

## Fig. 1
proliferation (left) and apoptosis (right) scores
image

## Fig. 2
Combined growth rate
image

## Fig. 3
clusters. Note proliferation cluster
image

Marius1311:
So depending on the initialization, we can end up with growth rates that are quite different. See Fig. 1 below for growth rates computed in 3 iterations, starting from uniform initialization and Fig. 2 for 3 iterations, starting from the score-based initial growth rate computed above. 

## Fig. 1
image

## Fig. 2
image

Marius1311:
Compare also transition matrices from uniform (Fig. 1) or score-based (Fig. 2) initialization. 

## Fig. 1
uniform
image

## Fig. 2
score-based
image

Marius1311 (approved):
This is fine from my side. Please review my suggested workflow and modify if needed, in particular, if you think it can be shortened. Otherwise, feel free to merge.

michalk8:
> I found the gene sets to compute the growth rates. See Fig. 1 for the proliferation and apoptosis scores and Fig. 2 for the combined computed growth rate. This matches annotations well, see Fig. 3, note the proliferation cluster gets higher growth rate.

Nice! Should we include these or not?

> Compare also transition matrices from uniform (Fig. 1) or score-based (Fig. 2) initialization.

Visually, they seem similar.

> This is fine from my side. Please review my suggested workflow and modify if needed, in particular, if you think it can be shortened. Otherwise, feel free to merge.

Ok, thanks, will take a look at this.

Marius1311:
For the gene sets, they will be specific to the organism (mouse, in this case), so I wouldn't include them. We can add a link to the tutorial again where they can be downloaded.

PR #607, opened 2021-06-08
Update docstring to manually set terminal states

### LINKED ISSUE #606: Update `estimator.set_terminal_states` docstring
- [ ] Other?

Outdated, based on approximate recurrent classes, should be macrostates, terminal states, etc. Also, unclear how to use pre-defined cluster-labels to define terminal states. For real-time data, would be nice to combine cluster labels with time annotations, to pick cells from a set of clusters, restricted to the last time points.

Marius1311:

## Title
Update the docstring for `estimator.set_terminal_states`. 

## Types of changes

## Description
Was outdated. @michalk8, if possible, could you please add the following two functionalities (additionally to what we currently have):
- allow dict's of the form `{'cluster_key_from_obs': ['cluster_1', 'cluster_2', ...]}, i.e. directly use a subset of pre-defined clusters to define terminal states
- if real-time annotations are present, add a mechanism such that terminal states can be restricted to come from a certain time-point (optionally). 

If the second point is too difficult, don't worry, then we don't do it. 

## Closes

closes #606

### DIFF
--- a/cellrank/tl/_utils.py
+++ b/cellrank/tl/_utils.py
@@ -906,15 +906,15 @@ def save_fig(
 
 
 def _convert_to_categorical_series(
-    rc_classes: Dict[Union[int, str], Sequence[Union[int, str]]], cell_names: List[str]
+    term_states: Dict[Union[int, str], Sequence[Union[int, str]]], cell_names: List[str]
 ) -> Series:
     """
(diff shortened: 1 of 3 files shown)


michalk8:
Tests pass, codecov still has problems. If it isn't fixed by the end of the week, I will start using token again, supposedly, it should work.

PR #615, opened 2021-06-20
Vein plot

### LINKED ISSUE #591: Add vein plot
- [x] New plotting function in `cr.pl`?
- [ ] Other?

This plot can nicely describe transition matrice., see fig. 7c from [Mittnenzweig. *et. al.*](https://www.cell.com/cell/fulltext/S0092-8674(21)00439-6).

michalk8:

## Types of changes

## Description

Add vein plot, see the below issue.

## How has this been tested?

TODO.

## Closes

closes #591 

## TODOs
0. [x] clean the code
1. [x] flow normalization option (currently, it's local - per time point, could/should be also global, across all)
2. [x] cluster order (default will be by incoming flow [either asc/desc], otherwise user specified), can also be used to visualize specific clusters only
3. [x] option to allow only specific timepoints (currently, only consecutive are used), or at least specify start/end time point only?
4. [x] remove clusters if they have no incoming flow visualized?
5. [x] in the original, they remove clusters without enough cumulative inc. flow, but I'd remove this option (currently, can filter which flow edges to visualize with `min_flow` + with 4., it can simulate the same effect
6. [x] add the vertical bar showing percentage as in the original?

@Marius1311 would be great if you could give it a try + some feedback regarding the last 3 points. On the lung data, you can run it as:
```python
k = ...
k.plot_flow("AT2 activated", "clusters", "day", dpi=200, min_flow=0.1)
```
and for any subclass of ExperimentalTimeKernel, you don't need to specify the time key ("day" in this case).
Here's how it looks right now:
index


Marius1311:
Very nice @michalk8! I will check this out today! Fantastic work, I'm really happy this worked out so quickly.

Marius1311:
One general question: is there are clever way to sort the rows, i.e. the clusters? Ideally, we would minimize the flow x the distance travelled, a bit similar to our circular projection.

Marius1311:
Can't call the method, for `wk.plot_flow(cluster_key='cell_sets', time_key='day')`, I get
```python
TypeError                                 Traceback (most recent call last)
<ipython-input-42-4ec506bf3aec> in <module>
----> 1 wk.plot_flow(cluster_key='cell_sets', time_key='day')

~/Projects/cellrank/cellrank/tl/kernels/_exp_time_kernel.py in plot_flow(self, cluster_key, time_key, *args, **kwargs)
    112         if time_key is None:
    113             time_key = self._time_key
--> 114         return super().plot_flow(cluster_key, time_key, *args, **kwargs)
    115 
    116     @property

TypeError: plot_flow() missing 1 required positional argument: 'time_key'
```

Marius1311:
Same result once I've converted my time point annotation to float, using `adata.obs['day'].cat.rename_categories(lambda x: float(x), inplace=True)`. I made sure this worked: `Categories (39, float64): [0.0 < 0.5 < 1.0 < 1.5 ... 16.5 < 17.0 < 17.5 < 18.0]`

Marius1311:
Can we include a convenient way of transition matrix caching that works for every kernel? I.e a way to save these to file, based on a path I optionally specify upon initialization?

michalk8:
> Can we include a convenient way of transition matrix caching that works for every kernel? I.e a way to save these to file, based on a path I optionally specify upon initialization?

All kernels have `.write(<path>)` method, which writes them to pickle (as well estimators + models). However, this also saves the AnnData object.

Marius1311:
> All kernels have `.write(<path>)` method, which writes them to pickle (as well estimators + models). However, this also saves the AnnData object.

What's the rationale behind this? I think it would make sense to cache only the most expensive computations, i.e. transition matrix computation.

michalk8:
> What's the rationale behind this? I think it would make sense to cache only the most expensive computations, i.e. transition matrix computation.

To easily reconstruct the objects for further analysis. Transition matrices are already being cached, although only in memory.

Marius1311:
Okay, I got it - we follow a different logic in scachepy though, which I think makes more sense - only caching expensive computations, to keep the analysis as interactive as possible while not writing too much to disk.

Marius1311:
Example: 
image
 
Issues: 
- x-axis -> return axis object?
- flow going backwards?
- allow cluster restriction and time-point restriction
- allow using the raw transition (transport) maps for the WOTKernel / subclass to also have this for future kernels

Multi-focal plot will be handled in a separate issue.

michalk8:
> x-axis -> return axis object?

Done via `show: bool = True` (however, the figure is still being shown even if `False`, will fix it later).

> flow going backwards?

Fixed.

> allow cluster restriction and time-point restriction

Fixed.

> allow using the raw transition (transport) maps for the WOTKernel / subclass to also have this for future kernels

Working on this.

Marius1311:
Nice, let me know you're done with everything and once you're happy with your results on the reprogramming data. Please do compare with Fig. 7g from Schiebinger et al., Cell 2019 to evaluate how well this visualization is working.

michalk8:
> Nice, let me know you're done with everything and once you're happy with your results on the reprogramming data. Please do compare with Fig. 7g from Schiebinger et al., Cell 2019 to evaluate how well this visualization is working.

Am satisfied with how it looks/works now, here's the figure (still need to fix the ticks):
index
Re visualizing transport maps - afraid it's not meaningful, since I have to normalize the flow - in the end, it ends up looking just like transition matrix.

Marius1311:
Thanks @michalk8, will check this out today!

Marius1311:
Comments
- description for `show` does not make sense: "If `True`, return :class:`matplotlib.pyplot.Axes`." Should be other way round, no? It returns if `False`
- logging is off I think: `Computing flow from `MEF/other` into `3` clusters in `26` time points
Plotting flow from `MEF/other` into `3` clusters in `26` time points`, for Fig. 1 below. However, I only see flow into 2 clusters, no?

Apart from these two *very minor* comments, this works great and can be merged!


## Fig. 1
called with 
```python
ax = wk.plot_flow(cluster_key='cell_sets', time_key='day', cluster='MEF/other', min_flow=0.1, xticks_step_size=4,
                  dpi=200, show=False, clusters=['MEF/other', 'MET', 'Stromal'],
                  time_points=days)

# prettify the plot a bit
locs, labels = plt.xticks() 
ax.set_xticks(locs)
ax.set_xticklabels(labels, rotation=90)

plt.show()
```
image

Marius1311 (approved):
See my two minor comments I posted - other than that, ready to be merged!

Marius1311:
We have to consider the naming of this function, since we also want a plotting function that can show the flow from multiple clusters. I suggest
- `plot_single_flow`
- `plot_multi_flow`

other suggestions?

Marius1311:
BTW, for the multi_flow plotting function, we can fight back the messyness by enforcing a tree-like structure: if cluster A gives rise to B and C, and C gives rise to D and E, then there can be no flow directly from A to the childs of C, i.e. no flow from A to D or E. That means we can fit something like a minimum spanning tree (MST) prior to computing the flows and regularize by that. Keep in mind that we do want to allow multi-furcations, i.e. A can have more than just 2 children.

michalk8:
> https://github.com/theislab/cellrank/pull/615#issuecomment-866668713

Should be both fixed.

> We have to consider the naming of this method, since we also want a plotting function that can show the flow from multiple clusters. I suggest

Agreed, it's now called `plot_single_flow`.

Thanks for the review @Marius1311 , I've added some tests, once they pass, I will be merging this.

PR #630, opened 2021-06-25
Prolif apop markers

### LINKED ISSUE #626: Add proliferation and apoptosis gene sets to estimate growth rates
- [ ] Other?

We need this for the external WOT kernel and for our real-time tutorial. @michalk8, could you prepare the abstraction/infrastructure for this please, I'll take care of finding the lists for the main model organisms.

michalk8:

## Types of changes

## Description

Add porlif/apop markers - @Marius1311 could you please add them?

## How has this been tested?

1 new test needs to be finished @Marius1311 .

## Closes

closes #626


michalk8:
@Marius1311 please thouroughly check if everything is as it should be (correct prolif. apop. markers for the 2 organism + their references).
For mouse, I've used the `Proliferation` and `Apoptosis` columns in [from WOT](https://www.cell.com/cms/10.1016/j.cell.2019.01.006/attachment/a687e497-b4c8-45a3-b786-b59db5f3b9d0/mmc2.xlsx) - the references are in the 2nd sheet - why are they human, though (am I missing something)?

Marius1311:
> @Marius1311 please thouroughly check if everything is as it should be (correct prolif. apop. markers for the 2 organism + their references).

Thanks, will have a look!

> For mouse, I've used the `Proliferation` and `Apoptosis` columns in [from WOT](https://www.cell.com/cms/10.1016/j.cell.2019.01.006/attachment/a687e497-b4c8-45a3-b786-b59db5f3b9d0/mmc2.xlsx) - the references are in the 2nd sheet - why are they human, though (am I missing something)?

They are using a paper that studies humans as a reference because they must have used orthologs - genes found in several species that go back to a common ancestor, see e.g. https://link.springer.com/referenceworkentry/10.1007%2F978-3-642-27833-4_1731-3. That means, say we have a human gene, called `ABCD`. Then, the ortholog would be the mouse gene `Abcd`. Note the difference in capitalization. This works for many genes, you just change the capitalization and you go from mouse to human. Note that this won't work for all genes, mice and humans *are* different in the end.

cellrank/external/kernels/_wot_kernel.py
```diff
@@ -149,21 +157,58 @@ def compute_initial_growth_rates(
...
+        that case, make sure to use several iterations in
+        :meth:`cellrank.external.kernels.WOTKernel.compute_transition_matris` by increasing the `growth_iters`
+        parameter. A value around 3 works well in most cases.
+
+        The markers used here were taken from the following sources:
+
+            - human, proliferation - :cite:`tirosh:16:science`.
+            - human, apoptosis - `Hallmark Apoptosis, MSigDB <https://www.gsea-msigdb.org/gsea/msigdb/cards/HALLMARK_APOPTOSIS>`_.
+            - mouse, proliferation - :cite:`tirosh:16:nature`.
+            - mouse, apoptosis - `Hallmark P53 Pathway, MSigDB <https://www.gsea-msigdb.org/gsea/msigdb/cards/HALLMARK_P53_PATHWAY>`_.
```
Marius1311, line 182: I would move this to another place - ideally, directly to the place where the gene sets are being defined. I can imagine other functions/kernels also requiring access to apoptosis /proliferation estimates, so we should make sure that we keep all the references in one common place.

Marius1311, line 182: I would move this to another place - ideally, directly to the place where the gene sets are being defined. I can imagine other functions/kernels also requiring access to apoptosis /proliferation estimates, so we should make sure that we keep all the references in one common place.

Marius1311 (approved):
Apart from my one comment, this is perfect and can be merged. We should just make sure that we don't scatter the description of our gene sets across the package, i.e. don't give their citations in the WOT kernel because there may be other kernels that also need them, and in that case, we want to have everything that relates to these gene sets in one place.

cellrank/external/kernels/_wot_kernel.py
michalk8: Currently, the class is not exposed through public API, nor shown in the docs. Unless you want to expose it, I'd say leave it as it is now.

Marius1311: okay, we can leave it for the release tomorrow, but as soon as we need these gene sets in other kernels, we should find a different solution.

PR #638, opened 2021-06-28
Speed up calculation of absorption probabilities

### LINKED ISSUE #619: Speed up the computation of absorption probabilities
- [ ] Other?

Currently, when we want to compute absorption probabilities towards terminal macrostates, we select the N most confidently assigned cells per macrostate (by default, N=30) and we compute absorption probabilities towards them individually. Afterwards, we sum up absorption probabilities across all cells assigned to the same macrostates, to get one number per macrosate. However, this can be massively speed up by creating one pseudo-cell per macrostate which summarizes these N cells and computing absorption probabilities towards just these. Ignoring parallelisation, that's a speed up by factor N, i.e. 30 by default. 

to do's include
- make sure there are regression tests for absorption probabilities before changing anything. If we don't have any, add at least one. 
- implement the new routine based on the pseudo-cells
- make sure tests (esp. regression tests) pass

WeilerP:

# Speed up calculation of absorption probabilities

## Types of changes

## Description

Changes the computation of absorption probabilities according to @Marius1311 suggestion in #619.

## How has this been tested?

Unit tests already exist. `tests/test_cflare.py::TestCFLARE.test_abs_probs_negative` needs to be updated.

## Closes

Closes #619.

### DIFF
--- a/cellrank/tl/estimators/_base_estimator.py
+++ b/cellrank/tl/estimators/_base_estimator.py
@@ -512,6 +512,15 @@ class BaseEstimator(LineageEstimatorMixin, Partitioner, ABC):
         q = t[trans_indices, :][:, trans_indices]
         s = t[trans_indices, :][:, rec_indices]
 
+        # take individual solutions and piece them together to get absorption probabilities towards the classes
+        macro_ix_helper = np.cumsum(
+            [0] + [len(indices) for indices in lookup_dict.values()]
+        )
+        s = np.concatenate(
+            [s[:, np.arange(a, b)].sum(axis=1) for a, b in _pairwise(macro_ix_helper)],
+            axis=1,
+        )
+
         # check for irreducibility
         if check_irreducibility:
             if self.is_irreducible is None:
@@ -525,7 +534,7 @@ class BaseEstimator(LineageEstimatorMixin, Partitioner, ABC):
         logg.debug(f"Found `{n_cells}` cells and `{s.shape[1]}` absorbing states")
 
         # solve the linear system of equations
-        mat_x = _solve_lin_system(
+        _abs_classes = _solve_lin_system(
             q,
             s,
             solver=solver,
@@ -557,18 +566,6 @@ class BaseEstimator(LineageEstimatorMixin, Partitioner, ABC):
             abs_time_means.index = self.adata.obs_names
         else:
             abs_time_means = None
-
-        # take individual solutions and piece them together to get absorption probabilities towards the classes
-        macro_ix_helper = np.cumsum(
-            [0] + [len(indices) for indices in lookup_dict.values()]
-        )
-        _abs_classes = np.concatenate(
-            [
-                mat_x[:, np.arange(a, b)].sum(1)[:, None]
-                for a, b in _pairwise(macro_ix_helper)
-            ],
-            axis=1,
-        )
 
         # for recurrent states, set their self-absorption probability to one
         abs_classes = np.zeros((len(self), n_macrostates))
--- a/tests/test_cflare.py
+++ b/tests/test_cflare.py
@@ -635,7 +635,7 @@ class TestCFLARE:
             return_value=abs_prob,
         )
 
-        with pytest.raises(ValueError, match=r"`1` value\(s\) are negative."):
+        with pytest.raises(ValueError, match=r"`2` value\(s\) are negative."):
             mc.compute_absorption_probabilities()
 
 


WeilerP:
@michalk8, `tests/test_cflare.py::TestCFLARE.test_abs_probs_negative` [here](https://github.com/theislab/cellrank/blob/f90ae2b2e87acf5d4f39492a803d060f8a688524/tests/test_cflare.py#L617) fails now since `cellrank.tl.estimators._base_estimator._solve_lin_system` now already returns the summed probabilities. Consequently, two values are now negative. I suggest to simply update the matching string accordingly. Or do you prefer a different solution?

michalk8:
> I suggest to simply update the matching string accordingly. 

Agree, thanks!

Marius1311:
Fantastic, thanks. Can we compare this in a unit test to previous, summed absorption probabilities? i.e. does the test you mention above do that on a test case problem @WeilerP?

WeilerP:
> Fantastic, thanks. Can we compare this in a unit test to previous, summed absorption probabilities? i.e. does the test you mention above do that on a test case problem @WeilerP?

Unit tests checking the correct calculation of absorption probabilities already exist (e.g. [here](https://github.com/theislab/cellrank/blob/f4a08e62f9edb4d3e97f82c6129ad083d43cfd04/tests/test_gpcca.py#L443)) and are still passing, i.e. the old and new implementations give the same result. The test I mentioned is only designed to trigger `ValueError` which circumvents the actual calculation of absorption probabilities to speed up the unit test.

Marius1311:
We just need some statement on the speedup now - that's why we did this in the first place. I would like to know whether we really get something in the range of 4x speedup on a medium sized dataset (~50k cells)

michalk8 (approved):

michalk8:
LGTM. Re failing macOS test: not sure what's happening on GA, but it seems slow - just ignore it for now.

Marius1311:
I'll also check this out today on the reprogramming data to see what speedup we get.

WeilerP:
> I'll also check this out today on the reprogramming data to see what speedup we get.

Cool. I decided to test it on the arabidopsis dataset (which is similar in size I guess).

Marius1311:
> > I'll also check this out today on the reprogramming data to see what speedup we get.
> 
> Nice! I decided to test it on the arabidopsis dataset (which is similar in size I guess).

yes! That's a good idea. Esp. for many terminal states, we should see significant speedup. If not, we have to check whether parallelization still does what it should. Original parallelization, @michalk8, was optimized to solve many linear solved, i.e. in the range of 60-150. With the new implementations, we will be looking at 2-10 linear solves, in most problems.

Marius1311:
I tested this on a subsampled version of the reprogramming data (41.5k cells) on my laptop for 4 terminal states by calling
`g.compute_absorption_probabilities(n_jobs=n_jobs, solver='gmres', use_petsc=True)`. I get the following results

- old runtime (8 cores): 8m 42sec
- new runtime (8 cores): 1m 29sec
- new runtime (4 cores): 1m 25sec
- new runtime (1 core): 1m 15sec

So, this is great. I take two things from this:
1. It worked! No matter how many cores I use, I get at least 5x speedup. 
2. For such a low number of linear systems, parallelisation actually does not help and it's faster using a single core. This probably changes when we consider more terminal states.

@WeilerP, from my point of view, this can be merged as soon as tests pass. What sort of speedup did you see in your tests? Similar?

Marius1311:
I should note one thing here: the reason why absorption probabilities take relatively long here is because I used the WOT kernel to compute the transition matrix. Unfortunately, solving an entropically-regularized OT problem leads to dense solutions, hence, the transition matrix has dense blocks on the 1st off-diagonal. Luckily, these are not that large because the number of cells per time point is reasonable. This becomes problematic for datasets with extremely large cell numbers distributed across few time-points - that would mean very large dense matrices which would massively slow down our linear solver. 

So, if I repeated the same experiment using e.g. the `VelocityKernel`, I would expect to see computation times much lower. In the CellRank paper, we compute absorption probs. for 100k cells on the server in 2 min (using the `VelocityKernel`).

michalk8:
Merging this (MacOS is expected to fail right now).

WeilerP:
@Marius1311, sorry for the delayed reply. I ran it on the [mouse gastrulation data](https://www.nature.com/articles/s41586-019-0933-9) containing ~90000 cells and 6 terminal states:

* old implementation (6 cores): 3:35min
* new impelementation (6 cores): 9sec
* new implementation (1 core): 28sec

So the new implementation is whopping 24 times (ish) faster.

> 2. For such a low number of linear systems, parallelisation actually does not help and it's faster using a single core. This probably changes when we consider more terminal states.

Just as a side remark: I would not expect an increase in performance using more than 4 cores since there were only 4 terminal states, i.e. only 4 processes to parallelize.

Marius1311:
Thanks for these results! This is really great, it's a fantastic speedup. Being able to process 90k cells in 9 seconds is just fantastic!

As to the number of cores, you are right, for 4 terminal states, it doesn't make sense to use more than 4 cores. I was doing this to see whether using more cores than terminal states, in which case we should prevent the user from doing this by setting the number of cores equal to the number of terminal states and issuing a warning. I ended up not doing it because the performance loss wasn't great, so I thought it doesn't really matter. Feel free to open a PR to fix this - it's minor but user-friendly.

michalk8:
>Just as a side remark: I would not expect an increase in performance using more than 4 cores since there were only 4 terminal states, i.e. only 4 processes to parallelize.
> As to the number of cores, you are right, for 4 terminal states, it doesn't make sense to use more than 4 cores. 

That's because n the parallelize function, you can't use more cores than jobs, I clip it.

Marius1311:
Nice. Would be good to issue a warning about this.

PR #641, opened 2021-06-28
Stacked histogram

### LINKED ISSUE #635: Add visualizations for macrostate composition
- [ ] Other?

After having computed macrostates, it's often very insightful to check which cells they're made up from, i.e. what the identity of the top `N` (by default, 30) cells is that have been assigned to any macrostate. For example, one may be interested in the distribution over clusters or time points (for time-series data). For an example plot, see Fig. 1 below, which shows the distribution over time-points. 

We should implement this as an estimator-level plotting method called `plot_macrostate_composition`.  Code for the actual plotting is given below. 

## Fig. 1
Screenshot 2021-06-28 at 11 48 14
Distribution over days. Note that the colors here aren't ideal. 

## Code
```python
def stacked_histogram(adata: AnnData, 
                      states_key: str = 'macrostates', 
                      cat_key: str ='cluster',
                      width: float = 0.8, 
                      figsize: Iterable[float] = (5, 4),
                      show_title: bool = True,
                      rotation: int = 0,
                      save: Optional[str] = None):
    """Plot stacked histogram for macrostats over categorical annotations
    
    Parameters
    ---
    adata
        Annotated data matrix
    states_key
        Key from `.obs` where macrostate clustering can be found
    cat_key
        Key from `.obs` where the categorical annotations can be found
    width
        Bar width
    figsize
        Size of the overall figure
    show_title
        Whether to display generated titles
    rotation
        x-label rotation
    save
        Path where the figure is to be saved
        
    Returns
    ---
    Noting, just plots
    """
    
    # pull annotations from AnnData object
    mask = ~adata.obs[states_key].isna()
    base_df = sc.get.obs_df(adata[mask], [cat_key, states_key])
    
    # define parameters
    macrostates = list(base_df[states_key].cat.categories)
    cats = list(base_df[cat_key].cat.categories)
    cats_colors = adata.uns[f"{cat_key}_colors"]
    ind = np.arange(len(macrostates))
    y_max = np.array([np.sum(base_df[states_key] == ms) for ms in macrostates]).max()
    
    # initialise the figure
    fig = plt.figure(None, figsize)
    bottom = np.array([0] * len(macrostates))

    # iterate over categories, plot one barchart per category
    for i, cat in enumerate(cats):
        frequencies = [np.sum(base_df.loc[base_df[states_key] == ms, cat_key] == cat) for ms in macrostates]
        plt.bar(ind, frequencies, width, label=cat, color=cats_colors[i], bottom=bottom)
        bottom += np.array(frequencies)
    
    # prettify the plot
    plt.ylabel('frequency')
    plt.xlabel('macrostate')
    plt.xticks(ind, macrostates)
    plt.xticks(rotation=rotation)
    
    if show_title:
        plt.title(f"distribution over `{cat_key}`")
    plt.ylim([0, y_max + 2])
    plt.legend(bbox_to_anchor=(1.05, 1), loc='upper left')
    
    if save is not None:
        plt.savefig(save)
    plt.show()
```

Note that I had to write annotations to AnnData before calling this funciton using
```
adata.obs['macrostates'] = g.macrostates
adata.uns['macrostates_colors'] = g.macrostates_memberships.colors
```

michalk8:

## Types of changes

## Description

Add `cr.tl.estimators.GPCCA.plot_macrostate_composition`.

## How has this been tested?

TODO.

## Closes

closes #635


Marius1311:
Very nice, see Fig. 1 below. Comments
- For the y-ticks, I would make sure that we always have one tick corresponding to the total cell number per macrostate, i.e. 30 in this case. I just expect to see a label corresponding to the number of cells per macrostate when looking into compositional questions. 
- the color issue has been fixed - well done @michalk8!
- what would happen if one macrostate had less than 30 cells assigned to it? I.e, think of the situation where we compute many macrostates in the pancreas - some will get less than 30 cells assigned to it because they are dominated by other states that "eat up" all of their cells. Would this throw an error? I guess the correct behaviour would just be for one of the stacks to be a bit smaller - can you check this one please? Should be possible by using the pancreas and computing many macrostates

Once that's done, this can be merged!

## Fig. 1
image

Marius1311 (approved):
See my posted comments, two minor things and then this is ready to be merged.

michalk8:
> For the y-ticks, I would make sure that we always have one tick corresponding to the total cell number per macrostate, i.e. 30 in this case. I just expect to see a label corresponding to the number of cells per macrostate when looking into compositional questions.

I've maded sure that we see the tick corresponding to the highest bar.

> the color issue has been fixed - well done @michalk8!

 Thanks.

> what would happen if one macrostate had less than 30 cells assigned to it? I.e, think of the situation where we compute many macrostates in the pancreas - some will get less than 30 cells assigned to it because they are dominated by other states that "eat up" all of their cells. Would this throw an error? 

Works as expected, see below:
hist

PR #660, opened 2021-07-06
Check value of frac_to_keep is admissible

### LINKED ISSUE #659: Bound `frac_to_keep` in hard thresholding pseudo-time scheme
- [ ] Other?

## Description

The fraction of neighbors to keep when using the hard thresholding scheme in the `PseudotimeKernel` should be in the interval `[0, 1]`.

WeilerP:
The value for `frac_to_keep` should fall in the interval `[0, 1]`. If
this is not the case, an error is thrown and the function exited.

## Types of changes

## Description

Checks that `frac_to_keep` lies within `[0, 1]`.

## Closes

Closes #659.

### DIFF
--- a/cellrank/tl/kernels/_pseudotime_kernel.py
+++ b/cellrank/tl/kernels/_pseudotime_kernel.py
@@ -105,9 +105,10 @@ class PseudotimeKernel(Kernel):
         Parameters
         ----------
         frac_to_keep
-            The `fract_to_keep` * n_neighbors closest neighbors (according to graph connectivities) are kept, no matter
+            The `frac_to_keep` * n_neighbors closest neighbors (according to graph connectivities) are kept, no matter
             whether they lie in the pseudotemporal past or future. This is done to ensure that the graph remains
-            connected. Only used when `threshold_scheme='hard'`.
+            connected. Only used when `threshold_scheme='hard'`. `frac_to_keep` needs to fall within the
+            interval `[0, 1]`.
         %(soft_scheme_kernel)s
         check_irreducibility
             Optional check for irreducibility of the final transition matrix.
--- a/cellrank/tl/kernels/_pseudotime_schemes.py
+++ b/cellrank/tl/kernels/_pseudotime_schemes.py
@@ -157,8 +157,9 @@ class HardThresholdScheme(ThresholdSchemeABC):
         n_neighs
             Number of neighbors to keep.
         frac_to_keep
-            The `fract_to_keep` * n_neighbors closest neighbors (according to graph connectivities) are kept, no matter
-            whether they lie in the pseudotemporal past or future.
+            The `frac_to_keep` * n_neighbors closest neighbors (according to graph connectivities) are kept, no matter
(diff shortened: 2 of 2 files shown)


WeilerP:
@michalk8, should we add a test checking that the error message is thrown when using inadmissible values for `frac_to_keep`?

cellrank/tl/kernels/_pseudotime_kernel.py
```diff
@@ -120,6 +120,12 @@ def compute_transition_matrix(
         :class:`cellrank.tl.kernels.PseudotimeKernel`
             Makes :attr:`transition_matrix` available.
         """
+        if frac_to_keep < 0 or frac_to_keep > 1:
```
michalk8, line 123: Remove this - it's not necessary (will be caught later in the PT scheme).

cellrank/tl/kernels/_pseudotime_schemes.py
```diff
@@ -164,6 +165,12 @@ def __call__(
         -------
         %(pt_scheme.returns)s
         """
+        if frac_to_keep < 0 or frac_to_keep > 1:
```
michalk8, line 168: Change it to this:
```python
if not (0 <= frac_to_keep <= 1):
    raise ValueError(f"Expected `frac_to_keep` to be in `[0, 1]`, found `{frac_to_keep}`.`")
```
I usually write error messages in this format (i.e. `Expected ..., found ...`).

michalk8 (changes requested):
Please mention that `frac_to_keep` needs to be in `[0, 1]` in the docstring as well e.g. as *Needs to be in `[0, 1]`.*
Also, there are 2 older typos (`fract_to_keep` instead of `frac_to_keep`) in the `compute_transition_matrix` of a kernel and `__call__` of a scheme.

cellrank/tl/kernels/_pseudotime_kernel.py
WeilerP: I added it in both places to receive the error as soon as possible and in case someone calls the `HardThresholdScheme` independently. But either way is fine by me. I removed it.

WeilerP:
@michalk8, any idea why `codecov` is failing? Also, all CI tests are shown to have passed but looking at the run [here](https://github.com/theislab/cellrank/actions/runs/1004344153), for example, it says that **test (macos-latest, 3.8, false, true)** failed. In the test logs, it encounters a `Segmentation fault`.

michalk8:
> @michalk8, any idea why codecov is failing? 

It's just status check saying whether diff target has been reached or not - I've kept the defaults, we can change it later [here](https://github.com/theislab/cellrank/blob/master/.coveragerc).

> test (macos-latest, 3.8, false, true) failed. In the test logs, it encounters a Segmentation fault.

I've sat macOS to allow failures, since in some cases, `jax` causes segfault on the CI (all the tests are related to `'stochastic'` mode in `VelocityKernel`). If you want to, you can try tackling this (but I don't think you'd be able to reproduce this locally).

Otherwise, LGTM, thanks @WeilerP .

PR #665, opened 2021-07-12
Fix `compute_projection` for kNN based kernels

### LINKED ISSUE #603: Embedding projection only works sometimes
The embedding projection of transition matrices via `kernel.compute_projection` produces unexpected results for some kernels. Our idea for this method was to have something similar to scVelo's visualization of velocities in a 2D embedding, i.e. a UMAP or t-SNE. So we re-implemented this and it works fine for the `VelocityKernel`, basically gives the same picture we get if we use scVelo for the plotting. 

To produce a projection of a transition matrix via the kernel interface, you just call
```python
kernel.compute_projection() 
scv.pl.velocity_embedding_stream(adata, vkey='T_fwd') # for streamlines, can use the other scVelo functions here as well
```
and this should work for any kernel. 

The trouble starts when we use kernels other than the `VelocityKernel`, see below for e.g. the external `WOTKernel`. In Fig. 1, I'm showing the trajectories, which look fine, they go from the early days (day 0) to the late days (day 18), where they terminate. However, when I'm computing the embedding projection, I get the arrows from Fig. 2. 

The same thing happens for the `PseudotimeKernel`, however, **only for the hard threshold scheme**, see Figs. 3 and 4 below for hard and soft thresholding schemes, respectively. For the soft threshold scheme, the projected arrows (streamlines) look fine and reflect the expected direction. For the hard threshold scheme, the projection is totally wrong. I would like to emphasize that **I don't think the actual transition matrix is wrong** - when I use the transition matrix to plot random walks, starting in the middle (Fev+ cluster) of the UMAP, they go into the right direction and terminate in the right clusters (e.g. Beta). See Fig. 5 for the simulated random walks. 

So something is going wrong here when we use the hard threshold scheme. Note that in the hard threshold scheme, we actually remove some graph edges whereas in the soft scheme, we just re-weigh them, but we always keep all graph edges. This means that the sparsity pattern in the adjacency matrix `A` is symmetric in the soft thresholding scheme, but not in the hard thresholding scheme. Similarity, the sparsity pattern is not symmetric in the WOT kernel, but it is symmetric in the Velocity Kernel. This could be part of the problem. 

Projection of the transition matrix into low dimensions isn't trivial, this has already been discussed in the original RNA Velocity publication (La Manno et al, Nature 2018) in [Supplementary Note 2](https://static-content.springer.com/esm/art%3A10.1038%2Fs41586-018-0414-6/MediaObjects/41586_2018_414_MOESM3_ESM.pdf), "Section 11. Uncertainty and limitations of neighborhood-based velocity projections". This would be a good starting point to look into this - I think we somehow have to be a bit careful with local cell density when projecting into low dimensions, but I don't really understand it yet. 

## Fig. 1
Trajectories on reprogramming data from Schiebinger et al., Cell 2019
image

## Fig. 2
Projected velocity streams for reprogramming data. 
image
Now these look very weird! I would expect the streamlines to point from earlier to later days. 

## Fig. 3
Pancreas data, soft threshold scheme
image

## Fig. 4
Pancreas data, hard threshold scheme
image

## Fig. 5
Pancreas data, simulated random walks
image
Starting points denoted by black dots, endpoint denoted by yellow dots. Showing 5 simulations. Random walks visualized through lines, colored according to simulation time from black (early) via purple to yellow (late). 

#### Versions:

> ...

WeilerP:
## Types of changes

### Bug fixes

* The embedding now uses all pairs of observations with non-trivial connectivities. As this requires a connectivity matrix, an `AttributeError` is thrown if the attribute `_conn` is not set.

### Changes

* Rename index in for loop from `i` to `row_id`
* Add warning if `_conn` is not defined, i.e. `None`

## Closes

Closes #603.


cellrank/tl/kernels/_base_kernel.py
```diff
@@ -294,21 +295,37 @@ def compute_projection(
                 "Compute transition matrix first as `.compute_transition_matrix()`."
             )
 
+        if self.kernels[0]._conn is None:
```
michalk8, line 298: I'd change this to something like to be more robust:
```python
kernels = [k for k in self.kernels if k._conn is not None]
if not len(kernels):
    ...
```

cellrank/tl/kernels/_base_kernel.py
```diff
@@ -294,21 +295,37 @@ def compute_projection(
...
+            raise AttributeError(
+                f"Connectivity matrix not defined for {self.kernels[0]}. As of now, "
+                "the embedding projection only works for kNN based kernels."
+            )
+
         start = logg.info(f"Projecting transition matrix onto `{basis}`")
         emb = _get_basis(self.adata, basis)
         T_emb = np.empty_like(emb)
 
+        T = self.transition_matrix.copy()
```
michalk8, line 308: Can you please rewrite this without using the copies? The code down below will be slightly more complicated,
but I think it's better (faster/memory-effiecient without copying).

cellrank/tl/kernels/_base_kernel.py
```diff
@@ -294,21 +295,37 @@ def compute_projection(
...
+
         with warnings.catch_warnings():
             warnings.simplefilter("ignore")
-            for i, row in enumerate(self.transition_matrix):
-                dX = emb[row.indices] - emb[i, None]
+            for row_id, row in enumerate(T):
+                conn_idxs = conn[row_id, :].indices
+                dX = emb[conn_idxs] - emb[row_id, None]
+
+                probs = np.zeros_like(conn_idxs, dtype=T.dtype)
```
michalk8, line 320: I'd move this in the `else` branch.

cellrank/tl/kernels/_base_kernel.py
```diff
@@ -294,21 +295,37 @@ def compute_projection(
...
+                probs[np.isin(conn_idxs, row.indices)] = row.data
```
michalk8, line 321: This assumes that `len(row.data) == np.sum(np.isin(conn_idxs, row.indices)])`, which imho needs not to hold.
Think `probs = row[:, conn_idxs].data` should do the same + might make it easier when not sorting indices/making a copy.

michalk8 (changes requested):

michalk8:
@WeilerP please also add 1 plotting test for the `compute_projection` using a `PT` kernel with hard thresholding scheme.

cellrank/tl/kernels/_base_kernel.py
WeilerP: Hm, yes I had only kNN based kernels in mind. There it should be okay, I think. We can do

```python
probs = np.zeros_like(conn_idxs, dtype=row.dtype)
overlap = np.isin(conn_idxs, row.indices)
probs[overlap] = row[:, conn_idxs[overlap]].data
```

but still need to sort, though, to make sure indices in `_conn` and `row` are ordered the same. `row[:, conn_idxs].data` subsets to the right data but returns them w.r.t. the indices of `row`.

Also

```python
probs = row[:, conn_idxs].data
```

would again only consider non-trivial transition probabilities, *i.e.*, the shape of `dX` and `probs` would mismatch for the `PseudoTimeKernel`.

michalk8: Your right, my mistake, should be `row[:, conn_idxs].A.squeeze()`. Here's my proposal (should be identical):
```python
            for row_id, row in enumerate(self.transition_matrix):
                conn_idxs = conn[row_id, :].indices
                dX = emb[conn_idxs] - emb[row_id, None]

                if np.any(np.isnan(dX)):
                    T_emb[row_id, :] = np.nan
                else:
                    probs = row[:, conn_idxs]
                    if issparse(probs):
                        probs = probs.A.squeeze()
                    dX /= np.linalg.norm(dX, axis=1)[:, None]
                    dX = np.nan_to_num(dX)
                    T_emb[row_id, :] = probs.dot(dX) - dX.sum(0) / dX.shape[0]
```
One problem I have with `row.sort_indices(); overlap = np.isin(conn_idxs, row.indices)` is that it relies on `.transition_matrix` (was done in https://github.com/theislab/cellrank/pull/586, but wasn't fixed for emb. projection), altough 99% of the time it will be sparse (based on types, not necessarily CSR).

WeilerP: Good point. I refactored the code as you suggested.

Marius1311:
Thanks @WeilerP! Do we need to be careful for the `WOTKernel`, since it's not KNN based and considering all neighbors in the embedding will add many unrelated cells? Two thoughts come to my mind:
- maybe this even helps for the WOT kernel?
- or it doesn't matter because the WOT projection currently doesn't work anyways

Any thoughts on this?

WeilerP:
> Thanks @WeilerP! Do we need to be careful for the `WOTKernel`, since it's not KNN based and considering all neighbors in the embedding will add many unrelated cells? Two thoughts come to my mind:
> 
> * maybe this even helps for the WOT kernel?
> * or it doesn't matter because the WOT projection currently doesn't work anyways
> 
> Any thoughts on this?

Yes, I'd say we need to be careful with WOT and cannot make any claim ATM. First of all, the kernel does not have a connectivity matrix by default so the new implementation will raise an `AttributeError`. If a connectivity matrix is added manually, the resulting embedding projection can already improve as the embedding projections [here](https://github.com/theislab/cellrank/issues/603#issuecomment-877067538) show. I'll open a separate issue for kernels not based on kNN as discussed in #603. There we can also investigate how combining `WOTKernel` w/ e.g. `ConnectivityKernel` looks/helps.

michalk8 (approved):
LGTM @WeilerP , thanks a lot, merging this.

PR #696, opened 2021-08-07
Add WOT thresholding

### LINKED ISSUE #686: Threshold WOT transition matrix
## Description

The WOT transition matrix is only block sparse slowing things down e.g. the Schur decomposition in GPCCA. It seems that many transition probabilities are pretty small s.t. we could optionally set all probabilities below a certain threshold to zero.

michalk8:

## Types of changes

## Description

Add ``threshold`` to :meth:`cellrank.external.WOTKernel.compute_transition_matrix` to increase sparsity/speed up GPCCA.

## How has this been tested?

TODO

## Closes

closes #686


Marius1311:
How does the user control this? By default, the thresholding should be turned off.

michalk8:
> How does the user control this? By default, the thresholding should be turned off.

It is.

WeilerP:
@michalk8, when I checked this on the Schiebinger reprogramming data, the transition matrix was not recalculated when setting a new threshold, i.e.

```python
adata = cr.datasets.reprogramming_schiebinger()
adata = adata[adata.obs["serum"] == "True"].copy()
sc.pp.subsample(adata, fraction=0.25, random_state=0)
adata

sc.pp.pca(adata)
sc.pp.neighbors(adata, random_state=0)

wk.compute_transition_matrix(growth_iters=3, growth_rate_key="growth_rate_init", last_time_point="connectivities")

wk.compute_transition_matrix(
    growth_iters=3, growth_rate_key="growth_rate_init", last_time_point="connectivities", threshold=1e-4
)

wk.compute_transition_matrix(
    growth_iters=3, growth_rate_key="growth_rate_init", last_time_point="connectivities", threshold=1e-2
)
```

did not recalculate the transition matrix in the last two steps but used the cached result, instead.

michalk8:
> @michalk8, when I checked this on the Schiebinger reprogramming data, the transition matrix was not recalculated when setting a new threshold, i.e.

Thanks for catching this, I forgot to add the threshold to params.

WeilerP:
@michalk8, I encountered the following bug: After plotting random walks, recomputing the transition matrix with `threshold` specified fails due to an `AttributeError`. I'm running the following code snippet

```python
import scanpy as sc
import cellrank as cr

adata = cr.datasets.reprogramming_schiebinger()
adata = adata[adata.obs["serum"] == "True"].copy()
sc.pp.subsample(adata, fraction=0.25, random_state=0)

sc.pp.pca(adata)
sc.pp.neighbors(adata, random_state=0)

wk = WOTKernel(adata, time_key="day")

wk.compute_initial_growth_rates(organism="mouse", key_added="growth_rate_init")
wk.compute_transition_matrix(growth_iters=3, growth_rate_key="growth_rate_init", last_time_point="connectivities")

wk.plot_random_walks(
    n_sims=300,
    max_iter=200,
    start_ixs={"day": 0.0},
    basis="force_directed",
    c="day",
    legend_loc="none",
    linealpha=0.5,
    dpi=150,
)

wk.compute_initial_growth_rates(organism="mouse", key_added="growth_rate_init")
wk.compute_transition_matrix(
    growth_iters=3, growth_rate_key="growth_rate_init", last_time_point="connectivities", threshold=1e-2
)
```

which throws

```pytb
AttributeError                            Traceback (most recent call last)
/var/folders/7p/j67nx0vs6j968c2w2d9pqmzryt4ryq/T/ipykernel_96203/3189190673.py in <module>
      1 wk.compute_initial_growth_rates(organism="mouse", key_added="growth_rate_init")
----> 2 wk.compute_transition_matrix(
      3     growth_iters=3, growth_rate_key="growth_rate_init", last_time_point="connectivities", threshold=1e-2
      4 )

~/code/cellrank/cellrank/external/kernels/_wot_kernel.py in compute_transition_matrix(self, cost_matrices, lambda1, lambda2, epsilon, growth_iters, solver, growth_rate_key, use_highly_variable, last_time_point, threshold, conn_kwargs, **kwargs)
    361             return self
    362 
--> 363         tmap = self._compute_pairwise_tmaps(
    364             adata,
    365             cost_matrices=cost_matrices,

~/code/cellrank/cellrank/external/kernels/_wot_kernel.py in _compute_pairwise_tmaps(self, adata, cost_matrices, solver, growth_rate_field, **kwargs)
    411                 *tpair, cost_matrix=cost_matrix
    412             )
--> 413             tmap.X = tmap.X.astype(np.float64)
    414             nans = int(np.sum(~np.isfinite(tmap.X)))
    415             if nans:

AttributeError: 'NoneType' object has no attribute 'X'
```

This error is also thrown when running the code snippet from above on `cellrank@dev`. In this case, I am confused why there is no error thrown due to the unexpected argument `threshold` (opened #721 for this).

michalk8:
@WeilerP
It looks like `self._ot_model.compute_transport_map` returned `None`, which can only happen [here](https://github.com/broadinstitute/wot/blob/ca5e94f05699997b01cf5ae13383f9810f0613f6/wot/ot/ot_model.py#L287-L292),
which can only happen because the `dtype` of the field in `adata.obs` has been changed (happens because you call `plot_random_walks` and https://github.com/theislab/scvelo/issues/612)

michalk8:
@WeilerP I've added methods for you to test (under `method` argument), options are:
- `perc-local` - per row percentile-based cut-off (`threshold` defines the percentile)
  note that in this case, if the row is uniformly distributed, it will remain so afterwards with the same number of elements
- `perc-global` - cut-off based off the whole transition matrix (`threshold` defines the percentile)
- `k` - keep only the top-k values (`threshold` defines the k)
- `fixed` - as before  (`threshold` defines itself)
- `auto` - largest value such that no completely 0 row happens  (`threshold` is ignored)

WeilerP:
@michalk8, @Marius1311 I checked the different methods:

## In a nutshell

The correlation between the absorption probabilities for the clipped and original transition matrices is high. In case of the percentile based thresholding, I could reach the 99-percentile as not all (compared to without thresholding) cell types were identified as macro-states anymore. Seems sub-optimal to me.
Computation time is comparable for all four approaches (~1-2 seconds GPCCA, ~1sec computing absorption probabilities).

## method = 'perc-local'

percentile_row_wise

## method = 'perc-global'

percentile_global

## method = 'k'

fixed_k

## method = 'auto'

|| Neural 1 | IPS | Trophoblast | Stromal |
| -------------| ------------- | --------------- | --------------- | --------------- |
|Correlation w/ original | 0.995 | 0.996 | 0.995 | 0.997 |

I noticed that for several cells (~20), this resulted in a single non-zero transition probability which seems a bit odd to me. What do you think?

## Final remarks

I'd either go with setting `k` manually, using `'auto'` or `'perc-global'`.

michalk8:
> I noticed that for several cells (~20), this resulted in a single non-zero transition probability which seems a bit odd to me. What do you think?

Don't think this is odd (I'd expected more just by pure coincidence).

I vote for the `perc-global` and `auto` approach (both disabled by default), since the former seems to perform
the best, the curve seems the smoothest (except for Trophoblast), it's more intuitive (I know by how much I reduce the matrix wheareas with `k` it's not obvious) and the `auto` is just convenient.
Let's discuss this tomorrow in the meeting; afterwards I will undo the last commit and use the approach we converged upon + add/update tests.

cellrank/external/kernels/_wot_kernel.py
```diff
@@ -499,6 +508,59 @@ def _generate_cost_matrices(
...
+            assert 0 <= threshold <= tmat.shape[1]
+            for i in range(tmat.shape[0]):
+                start, end = tmat.indptr[i], tmat.indptr[i + 1]
+                data = tmat.data[start:end]
+                ixs = np.argsort(data)[::-1]
+                data[ixs[threshold:]] = 0
+                tmat.data[start:end] = data
+        elif method == "auto":
+            minn = np.inf
+            for i in range(tmat.shape[0]):
```
WeilerP, line 541: Why not use list comprehension with a single `min`? `max` with `map`, converting to a list and then taking the minimum could potentially be even quicker but I doubt it (due to the list conversion)

WeilerP:
Used the `auto` threshold on the mouse gastrulation data from [Pijuana-Sala et al.](https://pubmed.ncbi.nlm.nih.gov/30787436/). Works like a charm and is lightning fast!
The time to compute the Schur decomposition dropped from ~17minutes to 38 seconds (**!!!**), computing macro states took only ~2 minutes instead of ~15:30 minutes!

cellrank/external/kernels/_wot_kernel.py
michalk8: Done.

PR #768, opened 2021-10-21
Added clarification on CytoTRACE kernel documentation

dawe:
This adds a clarification as suggested [here](https://github.com/theislab/cellrank/discussions/767#discussioncomment-1512858).
Also adds a paramter `top_n` to choose the number of top genes to be used to calculate the score. This is relevant for other kind of experiments (e.g. scATAC-seq) in which the number of fragments/regions to be considered may be different from the default used in RNA-seq.


Marius1311:
Thanks @dawe and @michalk8 !

PR #778, opened 2021-11-25
Add negative corr to CytoTRACE

michalk8:

## Description

Allow CytoTRACE to include negative correlations (``n_neg_genes``) and don't densify if aggregation is mean/median.
Also changes ``n_top_genes -> n_pos_genes``.

## How has this been tested?

Updated 1 test.

## Closes

N/A


Marius1311:
would it be possible to do this just via a weighted mean where we weigh each gene according to its correlation with gcs? I didn't fully think this though, just an idea.

michalk8:
> would it be possible to do this just via a weighted mean where we weigh each gene according to its correlation with gcs? I didn't fully think this though, just an idea.

Would say it's out of scope of this PR, though I've tried it and it correlates extremely highly (0.99985686). However, I can add a weight parameter to weight pos/negative score (currently it's 1:1).

Marius1311:
> > would it be possible to do this just via a weighted mean where we weigh each gene according to its correlation with gcs? I didn't fully think this though, just an idea.
> 
> Would say it's out of scope of this PR, though I've tried it (Zebrafish data) and it correlates extremely highly (0.99985686). However, I can add a weight parameter to weight pos/negative score (currently it's 1:1).

I would leave it as it is, 1:1 weighting is fine. Can you try this on one of the datasets I used to validate the re-implementation?

michalk8:
> I would leave it as it is, 1:1 weighting is fine. Can you try this on one of the datasets I used to validate the re-implementation?

Just tried it on the bone Marrow (pos is original, only top 200 most positively correlated genes, pos+neg includes also 200 most negatively correlated genes), here are the results:
marrow
Colored by GT CytoTRACE:
ct
`pos+neg` correlated with GT slightly worse (0.87854429 vs 0.91477446 for pos), with `pos` it correlates highly `0.94596034`.

michalk8:
@Marius1311 lmk if this can be merged.

Marius1311:
See my comments in MM where I send you the relevant NB to validate this.

michalk8:
> See my comments in MM where I send you the relevant NB to validate this.

Thanks and sorry for picking one actual example where CT doesn't work. Here are the results:
ng
pos

Marius1311:
Please add:
- diagonal line (dashed)
- what's the interpretation of this?

PR #791, opened 2021-12-16
Feature/refactor kernels

michalk8:
**IMPORTANT: Please search among the [Pull requests](../pulls) before creating one.**

## Description

Refactor kernel module. These changes include:
- allow kernel elementwise multiplication
- decouple CytoTRACE computation from kernel initialization, see :meth:`cellrank.tl.kernels.CytoTRACEKernel.compute_cytottrace`
- change argument ``mode`` to ``model`` in :class:`cellrank.tl.kernels.VelocityKernel`
- change argument ``scheme`` to ``similarity`` in :class:`cellrank.tl.kernels.VelocityKernel`
- remove ``'sampling'`` option in :class:`cellrank.tl.kernels.VelocityKernel`
- distinguish between unidirectional and bidirectional kernels
- bidirectional kernel inversion is no longer inplace, returns a new object
- add more memory efficient way of copying kernels
- improve :class:`cellrank.kernels.PrecomputedKernel`
- abstract kernel-related plotting functions (projection and random walks)
- temporarily remove kernel tricks examples
- fix various typos in docstrings
- fix corner-case bug when computing absorption probabilities
- fix not setting transport maps in :class:`cellrank.tl.kernels.TransportMapKernel`
- fix automatic threshold removing 1 row in :class:`cellrank.tl.kernels.TransportMapKernel` 

## How has this been tested?

Added a few new tests.

## Closes

closes #772

PR #804, opened 2022-02-01
Add order option in GPCCA.plot_coarse_T

michalk8:

## Description

Add ``order`` argument in :meth:`cellrank.tl.estimators.GPCCA.plot_coarse_T`.

## How has this been tested?

TODO
@Marius1311 for `incoming/outgoing`, I exclude the diagonal from the computation (for outgoing, it seems to sum to 1 so it's not interesting - not sure if this is always the case, maybe you know).
Also, for these 2 options, since off-diag elements can be negative, shall we take absolute value?

## Closes

discussed in #791


Marius1311:
Hi @michalk8, in short...
- Re excluding the diagonal: of course, that makes sense. You also include an option to sort by diagonal entries, don't you? 
- Re outgoing summing to one: that must hold for the matrix to represent the transition matrix of an embedded Markov chain
- Re negative elements: that's a difficult question - for now, let's not take the absolute value, let's add them up including their sign.

michalk8:
> Re excluding the diagonal: of course, that makes sense. You also include an option to sort by diagonal entries, don't you?

Yes.

> Re excluding the diagonal: of course, that makes sense. You also include an option to sort by diagonal entries, don't you?

Makes, sense, I've removed it since it's uninformative.

> Re negative elements: that's a difficult question - for now, let's not take the absolute value, let's add them up including their sign.

Ok, I've kept them.

PR #839, opened 2022-03-10
Change order to ascending when plotting macros

### LINKED ISSUE #836: Ordering of macrostates
Hi @michalk8, I'm looking into the new ordering methods you introduced for the `plot_coarse_T` method and I have a few remarks:
- could you please invert the sorting? Most users read figures from left to right, top to bottom. That means the early states, i.e. low stability, should be in the top left corner. Currently, it's the other way around, see Fig. 1 below. 
- this should be consistent across all ordering options - i.e. whatever sorting I use, I should always find my putative early states in the top left corner, and my putative late states in the lower right corner. It's also inverted for `stat_dist` at the moment, see Fig. 2.
- could you add add mode `outgoing` please? 

## Fig. 1
Screenshot 2022-03-06 at 15 16 04

## Fig. 2
Screenshot 2022-03-06 at 15 19 50

michalk8:

## How has this been tested?

Updated old figures.

## Closes

closes #836


Marius1311:
Thanks @michalk8

PR #920, opened 2022-07-02
[ci skip] Add example thumbnails

### LINKED ISSUE #903: Add Thumbnails for tutorials (and examples)
...
If you show me how to do this for one case @michalk8 I can do it for the others. Thanks!

michalk8:
This is the example @Marius1311 requested, see also here: https://nbsphinx.readthedocs.io/en/rtd-theme/subdir/gallery.html
I'd prefer to use the path in `_images`, since it is more robust than modifying the code cell's metadata and it doesn't create extra files if we were to manually place them in `_static/img`.

If you agree, please remove the last commit introducing the temporary `Untitled.ipynb` that was just for showcase and update `nbsphinx_thumbnails` accordingly.

closes #903

### DIFF
--- a/docs/source/Untitled.ipynb
+++ b/docs/source/Untitled.ipynb
@@ -1,0 +1,81 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "337faddc",
+   "metadata": {},
+   "source": [
+    "# Using `nbsphinx-thumbnail` metadata tag"
+   ]
+  },
+  {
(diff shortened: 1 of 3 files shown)


Marius1311:
Thanks for adding this, will get back to this over the coming days.

michalk8:
closing this because we will switch to `myst-nb`

Marius1311:
ok, but can we still have nice Thumbnails?

PR #943, opened 2022-09-05
Fix import issue with 'Iterable' from 'collections'

### LINKED ISSUE #942: cannot import name 'Iterable' from 'collections'
# Description
`Iterable` has moved from `collections` to `collections.abc` (see https://docs.python.org/3/library/collections.abc.html). Was deprecated in Python 3.9 and removed in Python 3.10 (https://docs.python.org/3/whatsnew/3.10.html#removed).

```python
from collections import Iterable
```

```pytb
ImportError: cannot import name 'Iterable' from 'collections'
```

Oisin-M:
'Iterable' has moved from 'collections' to 'collections.abc' since
Python 3.9

Resolves #942

## Description
Import `Iterable` from `collections.abc` and not `collections`. `Iterable` was removed from `collections` in Python 3.10

## How has this been tested?
Tested on Python 3.10 and 3.9

## Closes
#942

### DIFF
--- a/cellrank/pl/_heatmap.py
+++ b/cellrank/pl/_heatmap.py
@@ -5,7 +5,8 @@ import os
 from enum import auto
 from math import fabs
 from pathlib import Path
-from collections import Iterable, defaultdict
+from collections import defaultdict
+from collections.abc import Iterable
 
 from anndata import AnnData
 from cellrank import logging as logg


michalk8:
Thanks, looks like coverage upload is only failing, let's see if it resolves after merging, otherwise will take a look!

PR #987, opened 2023-02-15
[ci skip] first draft

MUCDK:
Draft of implementation of moscot kernel (only for those which output a transport matrix)

Replaces #986 and closes #982

PR #1016, opened 2023-03-31
`MoscotKernel`

MUCDK:
replaces #987


MUCDK:
I currently only test for `MoscotKernel.load` until we have agreed on how we want to enable instantiating it.

Moreover, there are some tests failling, but they are very cellrank-specific, so would appreciate your help @Marius1311 @WeilerP @michalk8 .

requirements.txt
```diff
@@ -1,5 +1,6 @@
 anndata>=0.8,<0.9
 docrep>=0.3.0
+git+https://github.com/theislab/moscot
```
michalk8, line 3: This cannot be here, please use just `moscot`.
Re failing tests: think they are coming from `scvelo`, try pinning `pandas<2` here temporarily.

requirements.txt
```diff
@@ -2,10 +2,11 @@ anndata>=0.8,<0.9
 docrep>=0.3.0
 joblib>=0.13.1
 matplotlib>=3.3.0
+moscot
```
michalk8, line 5: This should be an optional requirement.

requirements.txt
```diff
@@ -2,10 +2,11 @@ anndata>=0.8,<0.9
...
 networkx>=2.2
 numba>=0.51.0
 numpy>=1.17.0
-pandas>=1.3.0,!=1.4.0
+pandas==1.5.0
```
michalk8, line 9: <2, until `scvelo` is fixed.

tests/conftest.py
```diff
@@ -378,3 +378,47 @@ def test_matrix_4() -> np.ndarray:
     # fmt: on
 
     return p
+
+
+from moscot.datasets import simulate_data
```
michalk8, line 383: This will cause an issue if `moscot` is not installed, you might need to defined these fixtures differently.

tests/test_external.py
```diff
@@ -28,6 +28,15 @@ def _wot_not_installed() -> bool:
         return True
 
 
+def _moscot_not_installed() -> bool:
```
michalk8, line 31: This is also defined elsewhere, please remove one definition.

tests/test_moscot_interface.py
```diff
@@ -0,0 +1,178 @@
...
+    except ImportError:
+        return True
+
+
+moscot_not_installed_skip = pytest.mark.skipif(
+    _moscot_not_installed(), reason="moscot is not installed."
+)
+
+
+@moscot_not_installed_skip
```
michalk8, line 34: You can skip the whole file using `pytest` without the need to mark it like this.

tests/test_moscot_interface.py
```diff
@@ -0,0 +1,178 @@
...
+class TestMoscotKernel:
+    from moscot.problems import LineageProblem, TemporalProblem, SpatioTemporalProblem
```
michalk8, line 36: If you skip the whole file, this can be on top I think.

tests/test_moscot_interface.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def test_init_from_adata(self, adata_moscot: AnnData):
+        pass
```
michalk8, line 39: This test needs to be finished (or removed).

tests/test_moscot_interface.py
```diff
@@ -0,0 +1,178 @@
...
+
+    @pytest.mark.parametrize("backward", [True, False])
```
michalk8, line 41: I don't think we need to parametrize this by `backward`, esp. if you're not checking for it in the test.

cellrank/kernels/_moscot_kernel.py
```diff
@@ -0,0 +1,105 @@
...
+        """Load from a moscot model."""
+        return MoscotKernel(
+            adata=problem.adata,
+            problem=problem,
+            time_key=problem.temporal_key,
+            backward=backward,
+            **kwargs,
+        )
+
+    def compute_transition_matrix(
```
michalk8, line 42: I don't think there's a need to override this, just use the `_compute_tmap` API which should provide you with everything.

Regarding the `tmaps` argument, I think this is generic enough that it should be in the `TransportKernel` class, as a class method, e.g.,:
```python
@classmethod
def from_transport_maps(cls, adata: AnnData, tmaps: Dict[..., ...], **kwargs: Any) -> "TransportMapKernel":
    # do some checking whether all tmaps are present and synced with `adata`, etc.
    kernel = cls(adata, ...) 
    kernel._tmaps = tmaps
    kernel._is_precomputed = True  # needs to be set to false in `TransportMapKernel.__init__`
    return kernel
```

and then in `compute_transition_matrix`, you'd have:
```python
if not self._is_precomputed:
    self._tmaps = {}
    for t1, t2 in tqdm(timepoints, unit="time pair"):
        tmat = self._compute_tmap(t1, t2, **kwargs)
        self.transport_maps[t1, t2] = self._tmat_to_adata(t1, t2, tmat)
```

cellrank/kernels/_moscot_kernel.py
```diff
@@ -0,0 +1,105 @@
...
+    """Kernel for all moscot problems with a transport matrix as output."""
+
+    def __init__(
+        self, *args: Any, problem: Optional[Type[CompoundProblem]] = None, **kwargs: Any
+    ):
+        super().__init__(*args, **kwargs)
+        self._problem = problem
+
+    @classmethod
+    def load(
```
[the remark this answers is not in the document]
michalk8: In light of the other comment, if you have `from_transport_maps`, you can then have
```python
@classmethod
def from_moscot(cls, problem: "CompoundProblem", **kwargs: Any):
    tmaps = problem.solution.items()
    return cls.from_transport_maps(problem.adata, tmaps=tmaps, time_key=problem.time_key, **kwargs)
```

michalk8 (changes requested):

MUCDK:
I don't think we can simply override `_compute_tmap`, as we need different arguments in `compute_transition_matrix`. 
I.e., the signature of `MoscotKernel.compute_transition_matrix` should be along the lines of 
```python
def compute_transition_matrix(
        self,
        tmaps: Optional[Dict[Tuple[float, float], Type[BaseSolverOutput]]] = None,
        threshold: Optional[str] = None,
        self_transitions: Union[
            Literal["uniform", "diagonal", "connectivities", "all"],
            Sequence[float],
        ] = "connectivities",
        conn_weight: Optional[float] = None,
        conn_kwargs: Mapping[str, Any] = MappingProxyType({}),
        **kwargs: Any,
```

I.e. we need `tmaps` as additional argument compared to `TransportMapKernel.compute_transition_matrix()`

Issue #983, opened 2023-02-02
`MongeVelo` as a CR kernel


MUCDK:
As discussed, we should have a `MongeVelo` kernel in CR.

@WeilerP @soerenab let's discuss this.

michalk8:
closed via #1025

PR #1017, opened 2023-04-03
First draft of Monge Kernel

soerenab:
## Description
First draft of a kernel for velocities as computed by probabilistic Monge map (/Neural Optimal Transport). This kernel is necessary as Monge velocities cannot (generally) be used with the standard `VelocityKernel` due to shape mismatches: `VelocityKernel` expects velocities to be stored in an `adata.layer`, which requires the velocities to have the same shape as `adata.X`. However, Monge velocities may be computed in some dimensionally-reduced (latent) space so that the velocity dimensions do not match `adata.X.shape`. 
The `MongeKernel` subclasses the `VelocityKernel` and overrides `_read_from_adata()` to allow velocities to be read from `obsm`. Additionally, the kernel implements a simple wrapper for `compute_transition_matrix()` that calls `VelocityKernel`'s `compute_transition_matrix()`, but with the velocity model hard-coded as the deterministic model.

## How has this been tested?
So far only tested with a custom script, to see whether the code runs. Systematic tests still missing.

## Relevant issues:
https://github.com/theislab/cellrank/issues/983

### DIFF
--- a/cellrank/kernels/_monge_kernel.py
+++ b/cellrank/kernels/_monge_kernel.py
@@ -2,20 +2,13 @@ from typing import Any, Tuple, Union, Literal, Callable, Optional, Sequence
 
 from moscot.problems.time import TemporalNeuralProblem
 
-from anndata import AnnData
-from cellrank import logging as logg
 from cellrank._utils._docs import d, inject_docs
-from cellrank._utils._enum import _DEFAULT_BACKEND, Backend_t
-from cellrank.kernels.utils import Deterministic
-from cellrank.kernels.mixins import ConnectivityMixin
 from scvelo.preprocessing.moments import get_moments
-from cellrank.kernels._base_kernel import BidirectionalKernel
 from cellrank.kernels._velocity_kernel import VelocityKernel
-from cellrank.kernels.utils._similarity import Similarity, SimilarityABC
+from cellrank.kernels.utils._similarity import Similarity
 from cellrank.kernels.utils._velocity_model import BackwardMode, VelocityModel
 
 import numpy as np
-from scipy.sparse import issparse
 
 __all__ = ["MongeKernel"]
 
@@ -47,7 +40,7 @@ class MongeKernel(VelocityKernel):
             backward=backward,
             xkey=xkey,
             vkey=vkey,
-            attr=attr,  # this will be passed to _read_from_adata()
+            attr=attr,
             **kwargs,
         )
         self._logits: Optional[np.ndarray] = None
@@ -60,22 +53,17 @@ class MongeKernel(VelocityKernel):
         gene_subset: Optional[Union[str, Sequence[str]]] = None,
         **kwargs: Any,
     ) -> None:
-        super(VelocityKernel, self)._read_from_adata(
-            **kwargs
-        )  # TODO: Is there a better way?
+        super(VelocityKernel, self)._read_from_adata(**kwargs)
 
         if attr == "layers":
             self._xdata = self._extract_layer(xkey, subset=gene_subset)
             self._vdata = self._extract_layer(vkey, subset=gene_subset)
         elif attr == "obsm":
             self._xdata = self.adata.obsm[xkey]
-            self._vdata = np.array(
-                self.adata.obsm[vkey]
-            )  # TODO: I dont think we need self._vdata to be of type
-            # 'jaxlib.xla_extension.ArrayImpl', therefore I am converting to numpy array here. But not sure if this is
-            # the best way of doing so? Could velocities also be sparse? (jaxlib.xla_extension.ArrayImpl type causes
-            # errors in compute_transition_matrix() -> ModelABC.__init__() as jax' astype does not have keyword `copy`:
-            # Signature (dtype: Union[Any, str, numpy.dtype, jax._src.SupportsDType]) -> jax.Array
+            self._vdata = np.array(self.adata.obsm[vkey])
+            # NOTE: jaxlib.xla_extension.ArrayImpl type causes errors in
+            # compute_transition_matrix() -> ModelABC.__init__() as jax' astype does not have keyword `copy`.
+            # Converting to numpy array here to avoid this issue.
         else:
             raise ValueError(f"Invalid attr {attr}. Valid choices are {_ATTR_OPTIONS}")
 
@@ -97,11 +85,8 @@ class MongeKernel(VelocityKernel):
             Callable[[np.ndarray, np.ndarray, float], Tuple[np.ndarray, np.ndarray]],
         ] = Similarity.CORRELATION,
         softmax_scale: Optional[float] = None,
-        seed: Optional[
-            int
-        ] = None,  # TODO: does the seed impact results for the deterministic model?
         **kwargs: Any,
-    ) -> "VelocityKernel":
+    ) -> "MongeKernel":
         """
         Compute transition matrix based on velocity directions on the local manifold.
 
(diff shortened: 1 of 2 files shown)


cellrank/kernels/_monge_kernel.py
```diff
@@ -0,0 +1,116 @@
+from typing import Any, Tuple, Union, Literal, Callable, Optional, Sequence
+
+from moscot.problems.time import TemporalNeuralProblem
```
michalk8, line 3: There will need to be a try except statement, since we don't want to require `moscot` to be installed.

cellrank/kernels/_monge_kernel.py
```diff
@@ -0,0 +1,116 @@
...
+    ) -> None:
+        super(VelocityKernel, self)._read_from_adata(**kwargs)
+
+        if attr == "layers":
+            self._xdata = self._extract_layer(xkey, subset=gene_subset)
+            self._vdata = self._extract_layer(vkey, subset=gene_subset)
+        elif attr == "obsm":
+            self._xdata = self.adata.obsm[xkey]
+            self._vdata = np.array(self.adata.obsm[vkey])
+            # NOTE: jaxlib.xla_extension.ArrayImpl type causes errors in
```
michalk8, line 64: I think this comment should be removed and use `np.asarray(..., copy=False)` to avoid copying the data, if not necessary.

cellrank/kernels/_monge_kernel.py
```diff
@@ -0,0 +1,116 @@
...
+        attr: _ATTR_OPTIONS = "obsm",
+        xkey: Optional[str] = "X_pca",
+        vkey: Optional[str] = "Monge_velocities",
+        gene_subset: Optional[Union[str, Sequence[str]]] = None,
+        **kwargs: Any,
+    ) -> None:
+        super(VelocityKernel, self)._read_from_adata(**kwargs)
+
+        if attr == "layers":
+            self._xdata = self._extract_layer(xkey, subset=gene_subset)
```
michalk8, line 59: I think it might be better to generalize `_extract_layers` to also handle `attr` information rather than overriding `_read_from_adata` (lot's of code repeat itself here). Then in `VelocityKernel`, `attr` will not be exposed in `__init__`, only in `_read_from_adata` it will have the default `layers`.

cellrank/kernels/_monge_kernel.py
```diff
@@ -0,0 +1,116 @@
...
+        %(velocity_scheme)s
+        %(parallel)s
+        kwargs
+            Keyword arguments for the underlying the Deterministic VelocityModel.
+
+        Returns
+        -------
+        Self and updates :attr:`transition_matrix`, :attr:`logits` and :attr:`params`.
+        """
+        super().compute_transition_matrix(
```
michalk8, line 109: Think you can just do `return super().compute_transition_matrix(...)`

cellrank/kernels/_monge_kernel.py
```diff
@@ -0,0 +1,116 @@
...
+    ):
+        super().__init__(
+            tnp.adata,
+            backward=backward,
+            xkey=xkey,
+            vkey=vkey,
+            attr=attr,
+            **kwargs,
+        )
+        self._logits: Optional[np.ndarray] = None
```
michalk8, line 46: These should be defined already in `VelocityKernel`.

cellrank/kernels/_monge_kernel.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Description.
+
+    Args:
+        VelocityKernel (_type_): _description_
+    """
+
+    def __init__(
+        self,
+        tnp: TemporalNeuralProblem,
```
michalk8, line 31: I don't think we need to pass `TemporalNeuralProblem`, as you just extract the `adata` attribute (not sure whether it has `{x,v}keys` and `attr` you could extract from it, but if yes, in that case it would make more sense.
Which leaves the question - why is this problem class even needed, apart from setting certain defaults in `__init__`? Am fine with having a separate class, as long as, e.g., `compute_transition_matrix` does something "useful", like restricts which velocity modes are available (line right now).
Otherwise, would just refactor `_read_from_adata` to accept how velocity data can be stored (obsm) instead of defining this kernel, which would seems a bit redundant.

cellrank/kernels/_monge_kernel.py
```diff
@@ -0,0 +1,116 @@
...
+from scvelo.preprocessing.moments import get_moments
+from cellrank.kernels._velocity_kernel import VelocityKernel
+from cellrank.kernels.utils._similarity import Similarity
+from cellrank.kernels.utils._velocity_model import BackwardMode, VelocityModel
+
+import numpy as np
+
+__all__ = ["MongeKernel"]
+
+_ATTR_OPTIONS = Literal["obsm", "layers"]
```
michalk8, line 15: This is used only 2x (or 3x), but I think it's fair to use this type-hint in place rather than defining an alias.

cellrank/kernels/_monge_kernel.py
```diff
@@ -0,0 +1,116 @@
...
+            self._xdata = self._extract_layer(xkey, subset=gene_subset)
+            self._vdata = self._extract_layer(vkey, subset=gene_subset)
+        elif attr == "obsm":
+            self._xdata = self.adata.obsm[xkey]
+            self._vdata = np.array(self.adata.obsm[vkey])
+            # NOTE: jaxlib.xla_extension.ArrayImpl type causes errors in
+            # compute_transition_matrix() -> ModelABC.__init__() as jax' astype does not have keyword `copy`.
+            # Converting to numpy array here to avoid this issue.
+        else:
+            raise ValueError(f"Invalid attr {attr}. Valid choices are {_ATTR_OPTIONS}")
```
michalk8, line 68: Would follow this pattern to be consistent with the rest of the repo:
```python
"Invalid attribute `{attr!r}`. Valid options are: `'obsm', 'layers'`."
```

michalk8 (changes requested):

cellrank/kernels/_monge_kernel.py
soerenab: I am not sure I understand this sentence: "as you just extract the adata attribute (not sure whether it has {x,v}keys and attr you could extract from it, but if yes, in that case it would make more sense."

I think we really only need to take the `adata` that is stored in the `TemporalNeuralProblem` and `xkey`, `vkey` and `attr` only refer to the extracted `adata`, no other info in `TemporalNeuralProblem`.

soerenab: After reading your comment I realize that I previously also misunderstood the three different `model` options in the `VelocityKernel`'s `compute_transition_matrix()`: I thought these refer to different `scVelo` velocity models (deterministic, stochastic, dynamic or something like that) but they are really `cellrank` specific models of how to compute the transition matrix, right? In that case it doesn't actually make sense to hard-code to the deterministic model as I (currently implemented it) but to keep all options.

soerenab:
I will wait for Dominik's response (on whether he prefers to keep a separate kernel or whether we should discard this idea and instead extend the existing `VelocityKernel`) before implementing the suggested changes (Thanks, Mike!),

MUCDK:
> I will wait for Dominik's response (on whether he prefers to keep a separate kernel or whether we should discard this idea and instead extend the existing `VelocityKernel`) before implementing the suggested changes (Thanks, Mike!),

I think we can keep the `VelocityKernel` , it's a good idea, we just wanna make sure that we run the model beforehand and only read in the results. This means we cannot use the `velocityKernel` directly as it is now.

michalk8:
closed in favor of #1025

PR #1025, opened 2023-04-28
allow data and velocity to be read from obsm (as needed for MongeVelo…

soerenab:
…cities)

## Description
Changes in `_velocity_kernel.py` so that data can optionally be read from `obsm`. This is needed for MongeVelocities where the dimension of the velocities may differ from the dimension of `adata.X` so that they cannot be stored in `adata.layers` but only in `adata.obsm`. Defaults are set so that the new implementation matches the original implementation.

## How has this been tested?
So far I have only tested that 
```python
import numpy as np
import jax.numpy as jnp
import scanpy as sc
import scvelo as scv
import cellrank as cr
from cellrank.kernels import VelocityKernel
from moscot.datasets import simulate_data

adata = cr.datasets.pancreas_preprocessed()
scv.utils.show_proportions(adata)
vk = VelocityKernel(adata)
vk.compute_transition_matrix()
```
as well as 
```python
import numpy as np
import jax.numpy as jnp
import scanpy as sc
import scvelo as scv
import cellrank as cr
from cellrank.kernels import VelocityKernel
from moscot.datasets import simulate_data

adata = simulate_data(
    n_distributions=2,
    cells_per_distribution=40,
    n_genes=60,
    key="batch",
)

sc.pp.pca(adata, n_comps=30)
sc.pp.neighbors(adata)

from moscot.problems.time import TemporalNeuralProblem
tnp = TemporalNeuralProblem(adata)
tnp = tnp.prepare(time_key="batch", joint_attr = "X_pca")
tnp = tnp.solve(tau_a=1, tau_b=1, iterations=3)
source = jnp.array(adata.obsm["X_pca"].copy())
velocity = tnp[('0', '1')].solution.push(source) - source


tnp.adata.obsm["Monge_velocities"] = velocity

vk = VelocityKernel(adata, xkey="X_pca", vkey="Monge_velocities", attr="obsm")
vk.compute_transition_matrix()
```
run without error. Whether the output is correct has so far not been tested.

This PR replaces a older, separate PR which implemented a separate `MongeKernel`: https://github.com/theislab/cellrank/pull/1017 We realized that the `MongeKernel` is essentially identical to the `VelocityKernel` if we allow data to be read from `adata.obsm`.

### DIFF
--- a/cellrank/kernels/_velocity_kernel.py
+++ b/cellrank/kernels/_velocity_kernel.py
@@ -48,6 +48,7 @@ class VelocityKernel(ConnectivityMixin, BidirectionalKernel):
         backward: bool = False,
         xkey: Optional[str] = "Ms",
         vkey: Optional[str] = "velocity",
+        attr: Optional[Literal["layers", "obsm"]] = "layers",
         **kwargs: Any,
     ):
         super().__init__(
@@ -55,6 +56,7 @@ class VelocityKernel(ConnectivityMixin, BidirectionalKernel):
             backward=backward,
             xkey=xkey,
             vkey=vkey,
+            attr=attr,
             **kwargs,
         )
         self._logits: Optional[np.ndarray] = None
@@ -63,16 +65,31 @@ class VelocityKernel(ConnectivityMixin, BidirectionalKernel):
         self,
         xkey: Optional[str] = "Ms",
         vkey: Optional[str] = "velocity",
+        attr: Optional[Literal["layers", "obsm"]] = "layers",
         gene_subset: Optional[Union[str, Sequence[str]]] = None,
         **kwargs: Any,
     ) -> None:
         super()._read_from_adata(**kwargs)
 
-        if gene_subset is None and f"{vkey}_genes" in self.adata.var:
+        if attr not in ["layers", "obsm"]:
+            raise ValueError(
+                f"Invalid attribute `{attr!r}`. Valid options are: `'obsm', 'layers'`."
+            )
+        elif (
+            attr == "layers"
+            and gene_subset is None
+            and f"{vkey}_genes" in self.adata.var
+        ):
             gene_subset = self.adata.var[f"{vkey}_genes"]
+        elif attr == "obsm" and gene_subset is not None:
+            logg.warning(
+                "Found `gene_subset = {model!r}` but `gene_subset` is not supported for `attr = {model!r}`. "
+                "Setting `gene_subset = None`."
+            )
+            gene_subset = None
 
-        self._xdata = self._extract_layer(xkey, subset=gene_subset)
-        self._vdata = self._extract_layer(vkey, subset=gene_subset)
+        self._xdata = self._extract_data(key=xkey, attr=attr, subset=gene_subset)
+        self._vdata = self._extract_data(key=vkey, attr=attr, subset=gene_subset)
 
         nans = np.isnan(np.sum(self._vdata, axis=0))
         if np.any(nans):
@@ -239,16 +256,19 @@ class VelocityKernel(ConnectivityMixin, BidirectionalKernel):
         _, logits = model(n_jobs, backend)
         return 1.0 / np.median(np.abs(logits.data))
 
-    def _extract_layer(
+    def _extract_data(
         self,
         key: Optional[str] = None,
+        attr: Optional[Union[None, Literal["layers", "obsm"]]] = None,
         subset: Optional[Union[str, Sequence[str]]] = None,
         dtype: np.dtype = np.float64,
     ) -> np.ndarray:
         if key in (None, "X"):
             data = self.adata.X
-        elif key in self.adata.layers:
+        elif attr == "layers" and key in self.adata.layers:
             data = self.adata.layers[key]
+        elif attr == "obsm" and key in self.adata.obsm:
+            data = np.asarray(self.adata.obsm[key])
         else:
             raise KeyError(f"Unable to find data in `adata.layers[{key}]`.")
 
--- a/cellrank/kernels/_velocity_kernel.py
+++ b/cellrank/kernels/_velocity_kernel.py
@@ -76,11 +76,7 @@ class VelocityKernel(ConnectivityMixin, BidirectionalKernel):
     ) -> None:
         super()._read_from_adata(**kwargs)
 
-        if attr not in ["layers", "obsm"]:
-            raise ValueError(
-                f"Invalid attribute `{attr!r}`. Valid options are: `'obsm', 'layers'`."
-            )
-        elif (
+        if (
             attr == "layers"
             and gene_subset is None
             and f"{vkey}_genes" in self.adata.var
@@ -95,9 +91,11 @@ class VelocityKernel(ConnectivityMixin, BidirectionalKernel):
 
         self._xdata = self._extract_data(key=xkey, attr=attr, subset=gene_subset)
         self._vdata = self._extract_data(key=vkey, attr=attr, subset=gene_subset)
-        assert np.all(
-            self._xdata.shape == self._vdata.shape
-        ), f"Shape mismatch: {self._xdata.shape} vs {self._vdata.shape}"
+        assert np.testing.assert_array_equal(
+            x=self._xdata.shape,
+            y=self._vdata.shape,
+            err_msg=f"Shape mismatch: {self._xdata.shape} vs {self._vdata.shape}",
+        )
 
         nans = np.isnan(np.sum(self._vdata, axis=0))
         if np.any(nans):
--- a/cellrank/kernels/_velocity_kernel.py
+++ b/cellrank/kernels/_velocity_kernel.py
@@ -84,14 +84,14 @@ class VelocityKernel(ConnectivityMixin, BidirectionalKernel):
             gene_subset = self.adata.var[f"{vkey}_genes"]
         elif attr == "obsm" and gene_subset is not None:
             logg.warning(
-                "Found `gene_subset = {model!r}` but `gene_subset` is not supported for `attr = {model!r}`. "
-                "Setting `gene_subset = None`."
+                f"Found `gene_subset = {gene_subset!r}` but `gene_subset` is not supported for "
+                f"`attr = {gene_subset!r}`. Setting `gene_subset = None`."
             )
             gene_subset = None
 
         self._xdata = self._extract_data(key=xkey, attr=attr, subset=gene_subset)
         self._vdata = self._extract_data(key=vkey, attr=attr, subset=gene_subset)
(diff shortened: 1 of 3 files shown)


soerenab:
@MUCDK Is there a good way to check whether `vk.compute_transition_matrix()` computes reasonable results, e.g. do we have a reference (toy-) dataset to check this?

MUCDK:
To me this looks good, I would maybe add a simple test, once you try to read from `adata.layers`, and once you read from `adata.obsm` and check afterwards that it was read correctly. 

But for tests @michalk8 knows better.

cellrank/kernels/_velocity_kernel.py
```diff
@@ -63,16 +65,34 @@ def _read_from_adata(
...
         xkey: Optional[str] = "Ms",
         vkey: Optional[str] = "velocity",
+        attr: Optional[Literal["layers", "obsm"]] = "layers",
         gene_subset: Optional[Union[str, Sequence[str]]] = None,
         **kwargs: Any,
     ) -> None:
         super()._read_from_adata(**kwargs)
 
-        if gene_subset is None and f"{vkey}_genes" in self.adata.var:
+        if attr not in ["layers", "obsm"]:
```
michalk8, line 74: I think the condition can be simplified to and remove the warning + mention in the docstring of `gene_subset` it only available for `attr = 'layers'`.
```python
if attr == "layers" and gene_subset is None and f"{vkey}_genes" in self.adata.var:
   <as before>
self._x_data = ...
```

cellrank/kernels/_velocity_kernel.py
```diff
@@ -48,13 +48,15 @@ def __init__(
         backward: bool = False,
         xkey: Optional[str] = "Ms",
         vkey: Optional[str] = "velocity",
+        attr: Optional[Literal["layers", "obsm"]] = "layers",
```
michalk8, line 51: Missing docstring.

tests/test_kernels.py
```diff
@@ -700,6 +700,74 @@ def test_connectivities_key_kernel(self, adata: AnnData):
         np.testing.assert_array_equal(T_cr.A, adata.obsp[key])
 
 
+class TestVelocityKernelReadData:
```
michalk8, line 703: This test is not really necessary, as all the previous tests read from `attr = 'layers'`.
Would rather create 1 new tests where you can truncate velocities from `.layers` and put them in `obsm.`

tests/test_kernels.py
```diff
@@ -700,6 +700,74 @@ def test_connectivities_key_kernel(self, adata: AnnData):
...
+        gene_subset_reduced[10:] = False
+
+        vk = VelocityKernel(
+            adata,
+            xkey=xkey,
+            vkey=vkey,
+            attr="layers",
+            gene_subset=None,
+        )
+        assert np.all(
```
michalk8, line 719: np.testing.assert_array_equal (also in many other places).

tests/test_kernels.py
```diff
@@ -700,6 +700,74 @@ def test_connectivities_key_kernel(self, adata: AnnData):
...
+            vk_red._vdata
+            == adata.layers[vkey][:, np.asarray(gene_subset_reduced) & ~nans_v]
+        )
+        assert np.all(
+            vk_red._xdata
+            == adata.layers[xkey][:, np.asarray(gene_subset_reduced) & ~nans_v]
+        )
+
+    def test_read_correct_from_obsm(self, adata: AnnData):
```
michalk8, line 743: This test doesn't need to exist; the previous one should be parametrized using `@pytest.mark.parametrize("attr", ["layers", "obsm"])`

michalk8 (changes requested):

tests/test_kernels.py
soerenab: I think it makes sense to keep the warning tbh - and I will need an `elif` or `else` case anyways to set `gene_subset = None`

cellrank/kernels/_velocity_kernel.py
soerenab: I will need an `elif` or `else` in any case in order to set `gene_subset = None` when reading from obsm - and then I think it would make sense to keep the warning. Or is there any negative side effect that I am missing?

soerenab:
@michalk8 Thanks for your comments, I updated the code! Regarding the test: I followed your suggestion of a parameterized test which now checks reading from `layers` and `obsm` with and without specifying a `gene_subset`.

cellrank/kernels/_velocity_kernel.py
```diff
@@ -239,16 +262,19 @@ def _estimate_softmax_scale(
...
     ) -> np.ndarray:
         if key in (None, "X"):
             data = self.adata.X
-        elif key in self.adata.layers:
+        elif attr == "layers" and key in self.adata.layers:
             data = self.adata.layers[key]
+        elif attr == "obsm" and key in self.adata.obsm:
+            data = np.asarray(self.adata.obsm[key])
         else:
             raise KeyError(f"Unable to find data in `adata.layers[{key}]`.")
```
michalk8, line 279: f"Unable to find data in \`adata.{attr}[{key!r}]\`."

tests/test_kernels.py
```diff
@@ -700,6 +700,48 @@ def test_connectivities_key_kernel(self, adata: AnnData):
...
+            nans_v = np.isnan(np.sum(adata.layers[vkey], axis=0))
+        else:  # attr == "obsm"
+            xkey = "X_pca"
+            # setup reading from obsm: copy subset that matches X_pca shape to obsm
+            adata.obsm[vkey] = adata.layers[vkey][:, : adata.obsm[xkey].shape[1]]
+            nans_v = np.isnan(np.sum(adata.obsm[vkey], axis=0))
+
+        gene_subset = adata.var[f"{vkey}_genes"]
+        if use_gene_subset:
+            gene_subset[10:] = False
```
michalk8, line 724: would use
```python
else:
    gene_subset = None
```
and remove the extra if in `VelocityKernel`'s init.

cellrank/kernels/_velocity_kernel.py
```diff
@@ -31,13 +31,18 @@ class VelocityKernel(ConnectivityMixin, BidirectionalKernel):
...
-        Key in :attr:`anndata.AnnData.layers` where expected gene expression counts are stored.
+        Key in :attr:`anndata.AnnData.layers` or :attr:`anndata.AnnData.obsm` where expected gene expression counts are
+        stored.
     vkey
-        Key in :attr:`anndata.AnnData.layers` where velocities are stored.
+        Key in :attr:`anndata.AnnData.layers` or :attr:`anndata.AnnData.obsm` where velocities are stored.
     gene_subset
         List of genes to be used to compute transition probabilities.
         If not specified, genes from :attr:`anndata.AnnData.var` ``['{vkey}_genes']`` are used.
+        This feature is only available when reading from `anndata.AnnData.layers` and will be ignored otherwise.
```
michalk8, line 45: :attr:`anndata.AnnData.layers`

cellrank/kernels/_velocity_kernel.py
```diff
@@ -31,13 +31,18 @@ class VelocityKernel(ConnectivityMixin, BidirectionalKernel):
     ----------
     %(adata)s
     %(backward)s
+    attr
+        Attribute of `anndata.AnnData` to be read from. Should be `anndata.AnnData.layers` or `anndata.AnnData.obsm`.
```
michalk8, line 35: Would just just use:
```
Attribute of :class:`~anndata.AnnData` to read from.
```
as this should be pretty clear from the docs of `{x,v}key` where we read from.

cellrank/kernels/_velocity_kernel.py
```diff
@@ -48,13 +53,15 @@ def __init__(
         backward: bool = False,
         xkey: Optional[str] = "Ms",
         vkey: Optional[str] = "velocity",
+        attr: Optional[Literal["layers", "obsm"]] = "layers",
```
michalk8, line 56: Please move this argument below `backward` to maintain the order in the signature.

cellrank/kernels/_velocity_kernel.py
```diff
@@ -63,16 +70,32 @@ def _read_from_adata(
...
+            and gene_subset is None
+            and f"{vkey}_genes" in self.adata.var
+        ):
             gene_subset = self.adata.var[f"{vkey}_genes"]
-
-        self._xdata = self._extract_layer(xkey, subset=gene_subset)
-        self._vdata = self._extract_layer(vkey, subset=gene_subset)
+        elif attr == "obsm" and gene_subset is not None:
+            logg.warning(
+                f"Found `gene_subset = {gene_subset!r}` but `gene_subset` is not supported for "
```
michalk8, line 87:
```
f"Found `gene_subset != None`, but it is not supported for `adata.{attr}`. Using `gene_subset = None`"
```

tests/test_kernels.py
```diff
@@ -700,6 +700,48 @@ def test_connectivities_key_kernel(self, adata: AnnData):
...
 
+class TestVelocityKernelReadData:
+    @pytest.mark.parametrize("attr", ["layers", "obsm"])
+    @pytest.mark.parametrize("use_gene_subset", [True, False])
+    def test_read_correct_from_layers(
+        self,
+        adata: AnnData,
+        attr: str,
+        use_gene_subset: bool,
+        xkey: str = "Ms",
```
michalk8, line 711: Please remove `xkey/vkey` from the function signature and just hardcode them in the function.

tests/test_kernels.py
```diff
@@ -700,6 +700,48 @@ def test_connectivities_key_kernel(self, adata: AnnData):
         np.testing.assert_array_equal(T_cr.A, adata.obsp[key])
 
 
+class TestVelocityKernelReadData:
+    @pytest.mark.parametrize("attr", ["layers", "obsm"])
+    @pytest.mark.parametrize("use_gene_subset", [True, False])
+    def test_read_correct_from_layers(
+        self,
+        adata: AnnData,
+        attr: str,
```
michalk8, line 709: Literal["layers", "obsm"]

michalk8 (changes requested):

soerenab:
Thanks again for the feedback. I incorporated all your requests.

cellrank/kernels/_velocity_kernel.py
michalk8: It's just for consistency to maintain the same order in the signature as well as in the docs.

michalk8:
> Thanks again for the feedback. I incorporated all your requests.

Thanks a lot @soerenab , looks good. Tests are failing because of new `pandas` version and `scvelo`, the new tests are passing; merging this!

PR #1040, opened 2023-05-25
Rename `absorption` -> `fate`

### LINKED ISSUE #972: Naming conventions
Hi all, I have two naming suggestions. 
1. I would rename "lineages" to "trajectories" everywhere - we don't really estimate lineages, but abstracted cell trajectories
2. I would rename "absorption probabilities" as "fate probabilities" everywhere - since we want users to directly interact with estimators in version 2, I feel like the old term is too technical, and some users might not find what they are looking for (which will most likely be fate probabilities). 

What do both of you think? @michalk8 @WeilerP

michalk8:

## Description

TODO.

## How has this been tested?

No new tests.

## Closes

closes #972


Marius1311:
Amazing @michalk8, will take a look.

Marius1311 (changes requested):
Hi @michalk8, I checked this across all of the current tutorials I'm updating and the only issue I could find is this one: 

For some reason, the method to compute fate probabilities is currently called `g.compute_probabilities()`, it should be `g.compute_fate_probabilities`. Just `probabilities` is too generic. 

Apart from that, this is great and can be merged!

michalk8:
> For some reason, the method to compute fate probabilities is currently called g.compute_probabilities(), it should be g.compute_fate_probabilities. Just probabilities is too generic.

Thanks for checking this, fixed.

PR #1044, opened 2023-05-28
Feature/moscot kernel

michalk8:

## Description
.
Refactor `TransportMapKernel`.
@Marius1311 could you please try it on `moscot`? I've only tested this locally, should work, but expect some bugs. This is how you can initialize it:
```python
from cellrank.kernels import TransportMapKernel

problem = ...  # solved moscot problem

tmk = TransportMapKernel.from_moscot(problem)
```

## TODOs
- [x] implement `from_wot` method and remove `WOTKernel`
- [x] add code examples to the docs
- [x] polish documentation
- [x] add tests

## How has this been tested?

TODO.

## Closes

closes #1016


cellrank/kernels/_transport_map_kernel.py
```diff
@@ -229,80 +226,227 @@ def compute_transition_matrix(
...
+        )
+
+        if not problem.solutions:
+            raise RuntimeError(
+                "Problem contains no solutions. Please run `problem.solve(...)` first."
+            )
+
+        if isinstance(problem, TemporalMixin):
+            batch_key = problem.temporal_key
+        elif isinstance(problem, SpatialAlignmentMixin):
```
MUCDK, line 277: Is this neccesary? The `batch_key` in the `SpatioTemporalProblem` should be irrelevant, and be the same as the `time_key`.

michalk8: `batch_key` in the `SpatioTemporalKernel` will be the `time_key`, since it also subclasses the `TemporalMixin`, for which I check first.
Maybe the `batch_key` in the context of the kernel is not the best name, but couldn't find anything better. Maybe `covariate_key`? Pinging @Marius1311 for opinion.

MUCDK: Yeah doesn't make sense I think.
I mean it's a detail, but all moscot problems which make sense to use in CellRank, inherit from the `TemporalMixin` and hence have a `temporal_key`.

Marius1311: I agree, it does make make sense to allow this functionality for the`SpatialAligment` problem. The `TransportMap` Kernel should be coupled to a notion of time. CellRank is all about cellular dynamic, so we need a temporal element.

Marius1311:
Hi @michalk8, thanks for the great work, will check this out, hopefully later today!

Marius1311:
Hi Mike, currently, I struggle a bit with the dtypes for the temporal information - seems like moscot requires a numerical dtype, while the kernel wants a categorical - how should I go about this? Would be nice to have a bit of flexibility here, to allow a few different formats to be passed, as long as they can safely be converted internally.

Marius1311:
Okay, I played around a bit and couldn't find a single dtype for the time point column in AnnData that works for both moscot and the TransportMapKernel, one of the two always rises an exception.

Marius1311:
> > Okay, I played around a bit and couldn't find a single dtype for the time point column in AnnData that works for both moscot and the TransportMapKernel, one of the two always raises an exception.
> 
> In moscot we need numeric dtype for certain downstream tasks + `score_genes_for_marginals`

At least for `score_genes_for_marginals`, that should work as long as the categories have a numerical dtype, shound't it?

michalk8:
Should be fixed in https://github.com/theislab/moscot/pull/548

Marius1311:
Wonderful, thanks, I shall try this again.

Marius1311:
Thanks for merging this @michalk8! I will check this out soon.

PR #1062, opened 2023-06-21
Add banner

michalk8:
Add banner showing v1 -> v2


Marius1311:
I like this, great idea!

PR #1103, opened 2023-07-25
Update docs and README

### LINKED ISSUE #1102: Update documentation to include CellRank 2 preprint
## Description

With the CellRank 2 preprint out, we need to reference it in the documentation (how to get started, citation, etc.), README, etc.

WeilerP:

## Description

Updates README and documentation to include references to the CellRank 2 preprint.

## Closes

Closes #1102.


WeilerP:
@Marius1311, could you please update the conceptual overview to include metabolic labeling as well?

Marius1311 (approved):
Great, all of this works for me!

Marius1311:
the only thing I would change here @WeilerP: People need to realize first thing that this is the new version of CellRank - currently, it does not say "CellRank 2" anywhere prominently. I think this would be important to add. Otherwise, it's not entirely clear at first sight that something big has changed.

WeilerP:
> the only thing I would change here @WeilerP: People need to realize first thing that this is the new version of CellRank - currently, it does not say "CellRank 2" anywhere prominently. I think this would be important to add. Otherwise, it's not entirely clear at first sight that something big has changed.

Yes, good point. I updated the title in the README and index page to _CellRank 2: Unified fate mapping in multiview single-cell data_. What do you think?

Marius1311 (approved):
Great, I think that's good.

PR #1166, opened 2024-02-22
Add TSI score code

### LINKED ISSUE #1165: Add terminal state identification score
## Description

The _Terminal State Identification (TSI)_ score needs to be implemented through class method of GPCCA.

WeilerP:
## New

* Adds `GPCCA::get_tsi` to compute terminal state identification (TSI).
* Adds `GPPCA::get_tsi_score` to compute TSI score.
* Adds `GPCCA::plot_tsi` to visualize TSI.

## Closes

Closes #1165.


WeilerP:
What do you think, @michalk8?

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+        tsi_df = {
+            "Number of macrostates": [],
+            "Identified terminal states": [],
+            "Optimal identification": [],
+        }
+        for n_states, states in macrostates.items():
+            n_terminal_states = (
+                states.str.replace(r"(_).*", "", regex=True).drop_duplicates().isin(terminal_states).sum()
+            )
+            tsi_df["Number of macrostates"].append(n_states)
```
michalk8, line 569: I slightly prefer to not have spaces in the dataframe's columns and have all lower case, but no strong preference, so we can keep if you want.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+        n_macrostates
+            Maximum number of macrostates to consider.
+        terminal_states
+            List of terminal states.
+        cluster_key
+            Key in :attr:`~anndata.AnnData.obs` defining cluster labels including terminal states.
+
+        Returns
+        -------
+        Returns TSI as a Pandas DataFrame and adds the class attribute :attr:`tsi`.
```
michalk8, line 551: Would consider mentioning what the columns correspond to here.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+            raise RuntimeError("Compute TSI with `get_tsi` first.")
+
+        optimal_score = self.tsi["Optimal identification"].sum()
+
+        return self.tsi["Identified terminal states"].sum() / optimal_score
+
+    def plot_tsi(
+        self,
+        tsi_df: pd.DataFrame,
+        fname: Optional[PosixPath] = None,
```
michalk8, line 594: Optional[Path]

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+        x_offset: Optional[Tuple[float, float]] = None,
+        y_offset: Optional[Tuple[float, float]] = None,
+        **kwargs: Any,
+    ):
```
michalk8, line 598: Missing return type.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+        x_offset
+            Offset of x-axis. Defaults to `[0.2, 0.2]` if not specified.
+        y_offset
+            Offset of y-axis. Defaults to `[0.1, 0.1]` if not specified.
+        kwargs
+            Keyword arguments for :meth:`~seaborn.lineplot`.
+
+        Returns
+        -------
+        Returns TSI as a Pandas DataFrame and adds the class attribute :attr:`tsi`.
```
michalk8, line 616: Wrong docstring.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+        y_offset: Optional[Tuple[float, float]] = None,
+        **kwargs: Any,
+    ):
+        """Plot terminal state identificiation (TSI).
+
+        Parameters
+        ----------
+        tsi_df
+            Pre-computed TSI DataFrame.
+        fname
```
michalk8, line 605: We usually call it `save` and also include `dpi` + `figsize` in general (see other plotting methods, there's a `docrep` stuff for this as well), please use this.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+
+        optimal_score = self.tsi["Optimal identification"].sum()
+
+        return self.tsi["Identified terminal states"].sum() / optimal_score
+
+    def plot_tsi(
+        self,
+        tsi_df: pd.DataFrame,
+        fname: Optional[PosixPath] = None,
+        x_offset: Optional[Tuple[float, float]] = None,
```
michalk8, line 595: Use `x_offset: Tuple[float, float] = (0.2, 0.2)`, similarly for the `y_offset`.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+        y_offset: Optional[Tuple[float, float]] = None,
+        **kwargs: Any,
+    ):
+        """Plot terminal state identificiation (TSI).
+
+        Parameters
+        ----------
+        tsi_df
+            Pre-computed TSI DataFrame.
```
michalk8, line 604: Mention the method to compute here (or in the description) using :meth:\`get_tsi_score\`.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+        fname
+            File name under which the plot is saved. The plot is not saved if the argument is not specified.
+        x_offset
+            Offset of x-axis. Defaults to `[0.2, 0.2]` if not specified.
```
michalk8, line 608: Remove the mention of the defaults once they have been added to the signature; see the above comment (also for y-offset).

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+        handles[n_methods].set_linestyle("--")
+        handles = handles[: (n_methods + 1)]
+        labels = labels[: (n_methods + 1)]
+        fig.legend(handles=handles, labels=labels, loc="lower center", ncol=(n_methods + 1), bbox_to_anchor=(0.5, -0.1))
+        plt.tight_layout()
+        plt.show()
+
+        if fname is not None:
+            format = fname.suffix[1:]
+            fig.savefig(
```
michalk8, line 674: We have the `save_fig` function, why not use that?

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+                fname=fname,
+                format=format,
+                transparent=True,
+                bbox_inches="tight",
+            )
```
michalk8, line 679: Would consider returning the ax here.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+        """
+        if x_offset is None:
+            x_offset = [0.2, 0.2]
+
+        if y_offset is None:
+            y_offset = [0.1, 0.1]
+
+        optimal_identification = tsi_df[["Number of macrostates", "Optimal identification"]]
+        optimal_identification = optimal_identification.rename(
+            columns={"Optimal identification": "Identified terminal states"}
```
michalk8, line 626: Why is the data then saved before as `Optimal identification` if we always rename it here?

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+            y="Identified terminal states",
+            hue="Method",
+            style="line_style",
+            drawstyle="steps-post",
+            ax=ax,
+            **kwargs,
+        )
+
+        ax.set_xticks(df["Number of macrostates"].unique().astype(int))
+        for label_id, label in enumerate(ax.xaxis.get_ticklabels()):
```
michalk8, line 650: Consider using [MultipleTickLocator](https://matplotlib.org/stable/api/ticker_api.html#matplotlib.ticker.MultipleLocator)

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+        ax.get_legend().remove()
+
+        n_methods = len(df["Method"].unique())
+        handles, labels = ax.get_legend_handles_labels()
+        handles[n_methods].set_linestyle("--")
+        handles = handles[: (n_methods + 1)]
+        labels = labels[: (n_methods + 1)]
+        fig.legend(handles=handles, labels=labels, loc="lower center", ncol=(n_methods + 1), bbox_to_anchor=(0.5, -0.1))
+        plt.tight_layout()
+        plt.show()
```
michalk8, line 670: I wouldn't explicitly call `plt.show`

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+
+        ax.get_legend().remove()
+
+        n_methods = len(df["Method"].unique())
+        handles, labels = ax.get_legend_handles_labels()
+        handles[n_methods].set_linestyle("--")
+        handles = handles[: (n_methods + 1)]
+        labels = labels[: (n_methods + 1)]
+        fig.legend(handles=handles, labels=labels, loc="lower center", ncol=(n_methods + 1), bbox_to_anchor=(0.5, -0.1))
+        plt.tight_layout()
```
michalk8, line 669: This can be called in the figure level.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+
+        if y_offset is None:
+            y_offset = [0.1, 0.1]
+
+        optimal_identification = tsi_df[["Number of macrostates", "Optimal identification"]]
+        optimal_identification = optimal_identification.rename(
+            columns={"Optimal identification": "Identified terminal states"}
+        )
+        optimal_identification["Method"] = "Optimal identification"
+        optimal_identification["line_style"] = "--"
```
michalk8, line 629: Why is what seems a `seaborn` argument being passed to a dataframe? Shouldn't it be passed in the lineplot directly?

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+
+        df = tsi_df[["Number of macrostates", "Identified terminal states"]]
+        df["Method"] = self.kernel.__class__.__name__
+        df["line_style"] = "-"
+
+        df = pd.concat([df, optimal_identification])
+
+        fig, ax = plt.subplots(figsize=(6, 4))
```
michalk8, line 637: As mentioned before, I would allow for passing figsize/dpi.

michalk8 (changes requested):

michalk8 (commented):
Also please add tests.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -2,14 +2,16 @@
 import enum
 import pathlib
 import types
-from typing import Any, Dict, Literal, Mapping, Optional, Sequence, Tuple, Union
+from pathlib import PosixPath
```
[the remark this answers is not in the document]
WeilerP: Good point. Done!

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+            List of terminal states.
+        cluster_key
+            Key in :attr:`~anndata.AnnData.obs` defining cluster labels including terminal states.
+
+        Returns
+        -------
+        Returns TSI as a Pandas DataFrame and adds the class attribute :attr:`tsi`.
+        """
+        macrostates = {}
+        for n_states in range(1, n_macrostates):
```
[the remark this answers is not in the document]
WeilerP: Yes! I never encountered the problem because I had enough states computed to begin with.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+            tsi_df["Number of macrostates"].append(n_states)
+            tsi_df["Identified terminal states"].append(n_terminal_states)
+
+            if n_states <= max_terminal_states:
+                tsi_df["Optimal identification"].append(n_states)
+            else:
+                tsi_df["Optimal identification"].append(max_terminal_states)
+
+        tsi_df = pd.DataFrame(tsi_df)
+        self.tsi = tsi_df
```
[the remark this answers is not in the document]
WeilerP: Why not? I'd like to keep it s.t. we can easily plot the performance of different kernels in a single plot by passing the kernel to a dedicated function.

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: Done.

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: Eagle-eyed review - I like it!

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: Yes, I know. I honestly don't like it because it doesn't convey what the argument is meant for; the corresponding argument in `savefig` is also called `fname`. I changed it, though.

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: Lol, nice. I used `None` to not run into any issues with mutable types.

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: Because the column `"Identified terminal states"` already exists. It contains the number of states identified using the actual kernel.

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: Sure, I can do that. Related to the `save` argument: The function's argument is called differently again, namely `path`. We should change the code to consistently use `fname`, IMO.

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: No, the line style depends on the identification scheme. This way, seaborn uses a normal line for the kernel and a dashed line for the optimal identification.

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: It'd remove all ticks except for the fifth, right? I only want to remove the labels, not the ticks themselves.

WeilerP:
@michalk8, can you have another look, please? Once the code is fine, I'll add unit tests.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +535,140 @@ def set_initial_states(
...
+        handles, labels = ax.get_legend_handles_labels()
+        handles[n_methods].set_linestyle("--")
+        handles = handles[: (n_methods + 1)]
+        labels = labels[: (n_methods + 1)]
+        fig.legend(handles=handles, labels=labels, loc="lower center", ncol=(n_methods + 1), bbox_to_anchor=(0.5, -0.1))
+
+        if save is not None:
+            save_fig(fig=fig, path=save)
+
+        return fig, ax
```
michalk8, line 670: The ax is sufficient, the figure can be easily taken from it, so let's just return the `ax`.
Would consider also making it optional by adding `show: bool = False`, leave this up to you.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +535,140 @@ def set_initial_states(
...
+        Parameters
+        ----------
+        n_macrostates
+            Maximum number of macrostates to consider.
+        terminal_states
+            List of terminal states.
+        cluster_key
+            Key in :attr:`~anndata.AnnData.obs` defining cluster labels including terminal states.
+        kwargs
+            Keyword arguments passed to the class' `compute_macrostates` function.
```
michalk8, line 550: Use :meth:\`comptute_macrostates\`.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +535,140 @@ def set_initial_states(
         )
         return self
 
+    def get_tsi(self, n_macrostates: int, terminal_states: List[str], cluster_key: str, **kwargs: Any) -> pd.DataFrame:
+        """Compute terminal state identificiation (TSI).
+
```
michalk8, line 540: Also here, let's later add an explanation/link to the paper for reference.

src/cellrank/estimators/terminal_states/_gpcca.py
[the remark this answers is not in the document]
michalk8: Ok, but then it should be defined and types in `__init__`.

src/cellrank/estimators/terminal_states/_gpcca.py
michalk8: I understand, but it's not a big change, please consider using again `number_of_macrostates`, etc.); can be done easily with [DataFrame.rename](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rename.html).

src/cellrank/estimators/terminal_states/_gpcca.py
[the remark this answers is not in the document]
michalk8: I would still in this case return this in this case.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +534,150 @@ def set_initial_states(
...
+                tsi_df["Optimal identification"].append(n_states)
+            else:
+                tsi_df["Optimal identification"].append(max_terminal_states)
+
+        tsi_df = pd.DataFrame(tsi_df)
+        self.tsi = tsi_df
+
+        return tsi_df
+
+    def get_tsi_score(self) -> float:
```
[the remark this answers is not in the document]
michalk8: I think this function is fairly small, how about merging the functionality with `get_tsi` and giving it later the extra argument? I would prefer this than having 2 functions, as atm, it's less convenient.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +535,140 @@ def set_initial_states(
...
+        cluster_key
+            Key in :attr:`~anndata.AnnData.obs` defining cluster labels including terminal states.
+        kwargs
+            Keyword arguments passed to the class' `compute_macrostates` function.
+
+        Returns
+        -------
+        Returns TSI as a Pandas DataFrame and adds the class attribute :attr:`tsi`. The DataFrame contains the columns
+
+        - "Number of macrostates": Number of macrostates computed
```
michalk8, line 556: For consistency, the list should be formatted as:
```rst
- ``'col_name'`` - foo bar baz.
- ...
```

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +535,140 @@ def set_initial_states(
...
+        terminal_states
+            List of terminal states.
+        cluster_key
+            Key in :attr:`~anndata.AnnData.obs` defining cluster labels including terminal states.
+        kwargs
+            Keyword arguments passed to the class' `compute_macrostates` function.
+
+        Returns
+        -------
+        Returns TSI as a Pandas DataFrame and adds the class attribute :attr:`tsi`. The DataFrame contains the columns
```
michalk8, line 554: Would rephrase as:
```rst
Returns and updates :attr:`tsi`, a dataframe containing the following columns:

- ...
```

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +535,140 @@ def set_initial_states(
...
+        tsi_df: pd.DataFrame,
+        x_offset: Tuple[float, float] = (0.2, 0.2),
+        y_offset: Tuple[float, float] = (0.1, 0.1),
+        figsize: Tuple[float, float] = (6, 4),
+        dpi: Optional[int] = None,
+        save: Optional[Union[str, Path]] = None,
+        **kwargs: Any,
+    ) -> Tuple[plt.Figure, Axes]:
+        """Plot terminal state identificiation (TSI).
+
```
michalk8, line 603: Please add something along the lines:
```rst
See :meth:`get_tsi` on how to compute the TSI score.
```

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +535,140 @@ def set_initial_states(
...
+            raise RuntimeError("Compute TSI with `get_tsi` first.")
+
+        optimal_score = self.tsi["Optimal identification"].sum()
+
+        return self.tsi["Identified terminal states"].sum() / optimal_score
+
+    @d.dedent
+    def plot_tsi(
+        self,
+        tsi_df: pd.DataFrame,
```
michalk8, line 594: Why is `tsi_df` passed here and not being accessed from the class? If it's being saved, this method should access it.
In that case, please check `if self.tsi is None: raise RuntimeError(...)`.

michalk8 (changes requested):

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: I added a TODO.

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: Done.

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: Let's keep both. I don't see how you can extract the matplotlib figure instance from `ax`.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +536,165 @@ def set_initial_states(
         )
         return self
 
+    def tsi(
```
WeilerP, line 539: @michalk8, I refactored the function quite a bit:

1. It's called `tsi` now. I think it fits the remaining code base better since it is a function that is called by the user; we don't call methods `get_XXX`, for example, `compute_fate_probabilities` instead of `get_fate_probabilities`,  anywhere else as far as I am aware.
2. The arguments `terminal_states` and `cluster_key` are optional and only needed if the TSI needs to be (re)computed.
3. The TSI is saved as an AnnData object to check if it needs to be recomputed when passing a new set of terminal states or cluster key.

Let me know what you think.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +536,168 @@ def set_initial_states(
...
+            List of terminal states.
+        cluster_key
+            Key in :attr:`~anndata.AnnData.obs` defining cluster labels including terminal states.
+        kwargs
+            Keyword arguments passed to :meth:`compute_macrostates` function.
+
+        Returns
+        -------
+        Returns TSI score.
+
```
michalk8, line 563: Remove empty line.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +536,168 @@ def set_initial_states(
...
+                )
+                tsi_df["number_of_macrostates"].append(n_states)
+                tsi_df["identified_terminal_states"].append(n_terminal_states)
+
+                tsi_df["optimal_identification"].append(min(n_states, max_terminal_states))
+
+            tsi_df = AnnData(
+                pd.DataFrame(tsi_df), uns={"terminal_states": set(terminal_states), "cluster_key": cluster_key}
+            )
+            self._tsi = tsi_df
```
michalk8, line 597: Please add to the `__init__` method `self._tsi: Optional[AnnData] = None`.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +536,168 @@ def set_initial_states(
...
+            Offset of y-axis.
+        %(plotting)s
+        kwargs
+            Keyword arguments for :meth:`~seaborn.lineplot`.
+
+        Returns
+        -------
+        Plot TSI of the kernel and an optimal identification strategy.
+        """
+        if not hasattr(self, "_tsi"):
```
michalk8, line 636: Please use `if self._tsi is None` instead.

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +536,168 @@ def set_initial_states(
...
+        ----------
+        n_macrostates
+            Maximum number of macrostates to consider. Defaults to using all.
+        x_offset
+            Offset of x-axis.
+        y_offset
+            Offset of y-axis.
+        %(plotting)s
+        kwargs
+            Keyword arguments for :meth:`~seaborn.lineplot`.
```
michalk8, line 630: Use :func: instead of :meth:

src/cellrank/estimators/terminal_states/_gpcca.py
```diff
@@ -532,6 +536,168 @@ def set_initial_states(
...
+
+        Returns
+        -------
+        Plot TSI of the kernel and an optimal identification strategy.
+        """
+        if not hasattr(self, "_tsi"):
+            raise RuntimeError("Compute TSI with `tsi` first.")
```
michalk8, line 637: To retain the same style in the error messages, I'd use:
```python
raise RuntimeError("Compute TSI with `tsi` first as `.tsi()`.")
```

michalk8 (changes requested):

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: Yes, I thought the same, but I don't think the user needs access to the AnnData instance so I decided against it.

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: It's already there.

src/cellrank/estimators/terminal_states/_gpcca.py
WeilerP: Whoops. Thanks!

WeilerP:
@michalk8, do you happen to know why some plotting tests fail after I've added the unit tests?

Issue #1261, opened 2025-04-27
Invalid lineage name XYZ. Valid names are `[np.str_('4'), np.str_('0_1'), np.str_('0_2')]`.


Zethson:
I had disabled the ehrapy cellrank CI job for a while as it was broken and I wanted to wait for a proper release and not with development versions. Today, I tried enabling the job again and I see that my test notebook fails with:

```pytb
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[21], line 1
----> 1 drivers_1_1 = g.compute_lineage_drivers(lineages="1_1")
      2 adata.obs["fate_probs_1_1"] = g.fate_probabilities["1_1"].X.flatten()
      4 ep.pl.umap(
      5     adata,
      6     color=["fate_probs_1_1"] + list(drivers_1_1.index[:8]),
   (...)
     10     vmax="p96",
     11 )

File ~/miniconda3/envs/ehrapy/lib/python3.11/site-packages/cellrank/estimators/mixins/_lineage_drivers.py:134, in LinDriversMixin.compute_lineage_drivers(self, lineages, method, cluster_key, clusters, layer, use_raw, confidence_level, n_perms, seed, **kwargs)
    132     lineages = [lineages]
    133 if lineages is not None:
--> 134     _ = fate_probs[lineages]
    135 else:
    136     lineages = fate_probs.names

File ~/miniconda3/envs/ehrapy/lib/python3.11/site-packages/cellrank/_utils/_lineage.py:300, in Lineage.__getitem__(self, item)
    297     if isinstance(item, tuple):
    298         item = item[::-1]
--> 300 obj = self.__getitem(item)
    302 return obj.T if was_transposed else obj

File ~/miniconda3/envs/ehrapy/lib/python3.11/site-packages/cellrank/_utils/_lineage.py:383, in Lineage.__getitem(self, item)
    381             return self._mix_lineages(slice(None, None, None), item)
    382         if any(isinstance(i, str) for i in item):
--> 383             item = (slice(None, None, None), self._maybe_convert_names(item))
    384             col = item[1]
    386 shape, row_order, col_order = None, None, None

File ~/miniconda3/envs/ehrapy/lib/python3.11/site-packages/cellrank/_utils/_lineage.py:1028, in Lineage._maybe_convert_names(self, names, is_singleton, default, make_unique)
   1026                 name = default
   1027         else:
[...]

PR #1268, opened 2025-09-04
fix np str repr

Zethson:
- Deprecates Python 3.9 which is EOL. To be honest, I would jump immediately to 3.11+ (spec-0) or even 3.12+ (spec-0 in 2 months)
- Fixes https://github.com/theislab/cellrank/issues/1261


michalk8:
LGTM, thanks a lot!

Issue #1245, opened 2025-02-05
WARNING: Transition matrix does not exist for the given parameters. Recomputing


Enterprise-J:
Hi,

I received this warning when running `tmk.compute_transition_matrix(self_transitions = "all", threshold="auto", conn_weight=0.5)` following the RealTimeKernel tutorial, only changed `conn_weight=0.2` to `conn_weight=0.5`. What does that actually mean and what new parameters it will recompute with?

Thanks!

Marius1311:
Follow-up: the underlying bug @WeilerP spotted here (RealTimeKernel not recomputing when only `conn_weight` changes) was tracked in #1271 and is now fixed in #1333 — `conn_weight` is now part of the kernel's cache key and `params`, so changing it triggers a recomputation as expected.

PR #1333, opened 2026-06-04
fix(rtk): recompute transition matrix when conn_weight changes

### LINKED ISSUE #1271: RealTimeKernel does not recompute transition matrix
After computing the transition matrix with the RealTimeKernel, the kernel does not recompute it when changing `conn_weight`. I suspect the issue is that `tmk._params` only contains `self_transitions` and `threshold`, so the kernel uses the already computed/cached transition matrix automatically.

```python

# Compute transition matrix with different values for the `conn_weight` parameter
tmk = RealTimeKernel.from_moscot(tp)
tmk.compute_transition_matrix(self_transitions="all", conn_weight=0.2, threshold="auto")
old = tmk.transition_matrix.data.copy()

tmk.compute_transition_matrix(self_transitions="all", conn_weight=0.5, threshold="auto")

# Incorrectly asserts to True
np.testing.assert_almost_equal(old, tmk.transition_matrix.data)

# Compute transition matrix with different value for `conn_weight` using a separate kernel
tmk = RealTimeKernel.from_moscot(tp)
tmk.compute_transition_matrix(self_transitions="all", conn_weight=0.2, threshold="auto")

tmk_2 = RealTimeKernel.from_moscot(tp)
tmk_2.compute_transition_matrix(self_transitions="all", conn_weight=0.5, threshold="auto")

# Correctly asserts false
np.testing.assert_almost_equal(tmk.transition_matrix.data, tmk_2.transition_matrix.data)

# Compute transition matrix with same kernel object, but different values for the `conn_weight` parameter and set `conn_weight` in `_params` attribute
tmk = RealTimeKernel.from_moscot(tp)
tmk.compute_transition_matrix(self_transitions="all", conn_weight=0.2, threshold="auto")
tmk._params["conn_weight"] = 0.2
old = tmk.transition_matrix.data.copy()

tmk.compute_transition_matrix(self_transitions="all", conn_weight=0.5, threshold="auto")
#  Correctly asserts false
np.testing.assert_almost_equal(old, tmk.transition_matrix.data)
```

#### Versions:

> cellrank==2.0.7 scanpy==1.11.4 anndata==0.12.2 numpy==2.2.6 numba==0.61.2 scipy==1.16.2 pandas==2.3.2 pygpcca==1.0.4 scikit-learn==1.7.2 statsmodels==0.14.5 scvelo==0.3.3 pygam==0.9.1 matplotlib==3.10.6 seaborn==0.13.2

Marius1311:
## Description

`RealTimeKernel.compute_transition_matrix` did not include `conn_weight` in the parameters used to detect cache reuse (`cache_params` only held `threshold` and `self_transitions`). As a result, changing **only** `conn_weight` on an existing kernel object silently returned the stale, already-computed transition matrix instead of recomputing it. `conn_weight` was also missing from `kernel.params`.

This adds `conn_weight` to `cache_params`, so a changed value triggers a recomputation and is correctly recorded in `params`. This matches the reporter's diagnosis in #1271 (manually setting `tmk._params["conn_weight"]` worked around it).

## Changes

- `src/cellrank/kernels/_real_time_kernel.py`: add `conn_weight` to `cache_params`.
- `tests/test_kernels.py`: add `test_recompute_on_changed_conn_weight`, a regression test using explicit couplings (no moscot/wot dependency) asserting the transition matrix changes when only `conn_weight` is altered, and that `params["conn_weight"]` is recorded.

## Testing

`TestRealTimeKernel` passes (8 passed, 5 skipped — moscot/wot/R not installed).

Fixes #1271. Related to #1245.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/cellrank/kernels/_real_time_kernel.py
+++ b/src/cellrank/kernels/_real_time_kernel.py
@@ -182,6 +182,7 @@ class RealTimeKernel(UnidirectionalKernel):
         cache_params = dict(kwargs)
         cache_params["threshold"] = threshold
         cache_params["self_transitions"] = str(self_transitions)
+        cache_params["conn_weight"] = conn_weight
         if self._reuse_cache(cache_params):
             return self
 
--- a/tests/test_kernels.py
+++ b/tests/test_kernels.py
@@ -1212,6 +1212,27 @@ class TestRealTimeKernel:
             tgt_mask = (tmk.time == tgt).tolist()
             np.testing.assert_allclose(tmat[src_mask, :][:, tgt_mask].toarray(), expected[src, tgt])
 
+    def test_recompute_on_changed_conn_weight(self, adata_large: AnnData):
+        adata_large.obs["time"] = [0] * 50 + [1] * 50 + [2] * 50 + [3] * 50
+        adata_large.obs["time"] = col = adata_large.obs["time"].astype("category")
+        cats = col.cat.categories
+
+        rng = np.random.default_rng(0)
+        couplings = {}
+        for src, tgt in zip(cats[:-1], cats[1:]):
+            n, m = np.sum(col == src), np.sum(col == tgt)
+            couplings[src, tgt] = np.abs(rng.normal(size=(n, m)))
+
+        tmk = RealTimeKernel(adata_large, time_key="time", couplings=couplings)
+        tmk = tmk.compute_transition_matrix(self_transitions="all", conn_weight=0.2, threshold="auto")
+        old = tmk.transition_matrix.copy()
+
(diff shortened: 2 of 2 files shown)


Marius1311:
Update on testing: installed `moscot` and `wot` locally, so the full `TestRealTimeKernel` suite now runs with no skips — **13 passed**, including `test_from_moscot` (sequential + triu policies and the spatiotemporal problem) and `test_from_wot`. The fix is therefore verified through the real moscot OT path that matches the original reproducer, not just synthetic explicit couplings.

